PR #1, opened 2023-07-06
Use scikit-build to build and link extension

brandon-b-miller:
This PR converts the build process to use scikit-build which uses cmake under the hood to link the final python extension wrapping the `nvJitLink` c++ library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,11 @@ add_library(_nvjitlinklib MODULE nvjitlink/_nvjitlinklib.cpp)
 python_extension_module(_nvjitlinklib)
 
 target_link_libraries(_nvjitlinklib CUDA::nvJitLink_static CUDA::nvptxcompiler_static)
-target_compile_options(_nvjitlinklib PRIVATE -Werror -Wall -O0)
+
+if(WIN32)
+  target_compile_options(_nvjitlinklib PRIVATE -Werror -Wall -/Od)
+else()
+  target_compile_options(_nvjitlinklib PRIVATE -Werror -Wall -O0)
+endif()
 
 install(TARGETS _nvjitlinklib LIBRARY DESTINATION nvjitlink)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,27 @@
+cmake_minimum_required(VERSION 3.23.1 FATAL_ERROR)
+
+enable_language(C CXX)
+
+include(fetch_rapids.cmake)
+
+include(rapids-cmake)
+include(rapids-cpm)
+include(rapids-find)
+
+rapids_cpm_init()
+
+rapids_find_package(
+  CUDAToolkit 12.2 REQUIRED
+)
+
+project(nvjitlink)
+
+find_package(PythonExtensions REQUIRED)
+
+add_library(_nvjitlinklib MODULE nvjitlink/_nvjitlinklib.cpp)
+python_extension_module(_nvjitlinklib)
+
+target_link_libraries(_nvjitlinklib CUDA::nvJitLink_static CUDA::nvptxcompiler_static)
+target_compile_options(_nvjitlinklib PRIVATE -Werror -Wall -O0)
+
+install(TARGETS _nvjitlinklib LIBRARY DESTINATION nvjitlink)
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,8 @@ setup(
     name='nvjitlink',
     version='0.0.0',
     description='nvJitLink Python binding',
-    packages=['nvjitlink'],
-    python_requires=">=3.9"
+    packages=['nvjitlink', 'nvjitlink.tests'],
+    package_data={'nvjitlink.tests': ['test_device_functions.*', 'undefined_extern.cubin']},
+    python_requires=">=3.9",
+
 )
--- a/setup.py
+++ b/setup.py
@@ -12,88 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
-import os.path
-import shutil
-import sys
-from distutils.command.build_ext import build_ext
-from distutils.sysconfig import get_config_var, get_python_inc
-from setuptools import setup, Extension
-
-include_dirs = [os.path.dirname(get_python_inc())]
-library_dirs = [get_config_var("LIBDIR")]
-
-# Find and add CUDA include paths
-CUDA_HOME = os.environ.get("CUDA_HOME", False)
-if not CUDA_HOME:
-    path_to_cuda_gdb = shutil.which("cuda-gdb")
-    if path_to_cuda_gdb is None:
-        raise OSError(
-            "Could not locate CUDA. "
-            "Please set the environment variable "
-            "CUDA_HOME to the path to the CUDA installation "
-            "and try again."
-        )
-    CUDA_HOME = os.path.dirname(os.path.dirname(path_to_cuda_gdb))
-if not os.path.isdir(CUDA_HOME):
-    raise OSError(f"Invalid CUDA_HOME: directory does not exist: {CUDA_HOME}")
-include_dirs.append(os.path.join(CUDA_HOME, "include"))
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+cmake_minimum_required(VERSION 3.23.1 FATAL_ERROR)
+
+enable_language(C CXX)
```
vyasr, line 3: This should be specified in the project call.

CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+include(rapids-cpm)
+include(rapids-find)
+
+rapids_cpm_init()
+
+rapids_find_package(
+  CUDAToolkit 12.2 REQUIRED
+)
+
+project(nvjitlink)
```
vyasr, line 17: Put the languages here. Should also specify a version.

CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+cmake_minimum_required(VERSION 3.23.1 FATAL_ERROR)
```
vyasr, line 1: [nvptxcompiler requires at least 3.25](https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html#nvptx-compiler). You may want to just require 3.26.4 since that's what RAPIDS is doing now.

setup.py
vyasr, line 1: As a tiny library starting almost from scratch, I might recommend trying to use `scikit-build-core` from the get-go instead of `scikit-build` (unless there are plans to start building wheels for this package in 23.08 since then I assume you'd be using the shared workflows for the rest of RAPIDS and those make scikit-build-specific assumptions).

vyasr (changes requested):

CMakeLists.txt
bdice:
```suggestion
cmake_minimum_required(VERSION 3.26.4 FATAL_ERROR)
```

CMakeLists.txt
bdice:
```suggestion
project(
  nvjitlink
  LANGUAGES C CXX
)
```

setup.py
vyasr: setup.py is a setuptools thing. Switching to `scikit-build-core` implies removing `setup.py` altogether.

CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+include(fetch_rapids.cmake)
+
+include(rapids-cmake)
+include(rapids-cpm)
+include(rapids-find)
+
+rapids_cpm_init()
+
+rapids_find_package(
+  CUDAToolkit 12.2 REQUIRED
```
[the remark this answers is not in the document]
gmarkall: We are quite likely to always want to build with the very latest CUDA, because the aim is to support PTX coming from any released version of the CUDA toolkit.

setup.py
gmarkall: I like the idea of trying `scikit-build-core` in principle. In practice having an editable install seems to have a few caveats: https://scikit-build-core.readthedocs.io/en/latest/configuration.html#editable-installs - this isn't a problem for the build in general, but I use an editable install (or whatever  `setup.py develop` should be called) as my usual workflow - @vyasr have you used `scikit-build-core` with an editable install before? Or, what's a good workflow to use with it?

vyasr: I've tested it once or twice, but I haven't used it with any library in production. The workflow that worked was just a `pip install -e` instead of `python setup.py develop`.

I will say that while it does have caveats, those limitations (and more) are all present (and often in worse form) when using scikit-build classic. It can easily surprise you what gets rebuilt or not rebuilt when using those. Since you have a CPython extension here, do you benefit much from editable installs? You have to keep rebuilding from source every time unless you only change pure Python, right? For those cases scikit-build-core's support should work out of the box as well as the current support in classic.

brandon-b-miller: I made the switch here, let me know if things look ok.

CMakeLists.txt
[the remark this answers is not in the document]
vyasr: So the trick is going to be determining what the latest CUDA to support in different distribution mechanisms is. For pip wheels it may be that we always want to use the latest released CUDA since that's always installable, whereas for conda it may be lagged based on when we update conda packages to support the latest CUDA. IOW I don't know if it makes sense to encode a specific minor version here, but rather rely on that being set correctly in the appropriate build scripts for different distributions (e.g. meta.yaml recipes for conda or containers used for pip wheel builds). We'll have to maintain that second set of sources anyway, so we may as well make those the source of truth.

CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+find_package(
+  CUDAToolkit 12.0 REQUIRED
+)
+target_link_libraries(_nvjitlinklib PRIVATE CUDA::nvJitLink_static CUDA::nvptxcompiler_static)
+
+if(WIN32)
+  target_compile_options(_nvjitlinklib PRIVATE -Werror -Wall -/Od)
+else()
+  target_compile_options(_nvjitlinklib PRIVATE -Werror -Wall -O0)
+endif()
```
vyasr, line 22: Maybe cleaner to do
```suggestion
set(opt_option -O0)

if(WIN32)
  set(opt_option -/Od)
endif()
target_compile_options(_nvjitlinklib PRIVATE -Werror -Wall ${opt_option})
```

CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+cmake_minimum_required(VERSION 3.26.4 FATAL_ERROR)
+
+project(
+  nvjitlink
+  VERSION 0.1.0
+  LANGUAGES C CXX
```
vyasr, line 6: Do we need C or is it just C++ code?
```suggestion
  LANGUAGES CXX
```

pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[tool.scikit-build]
+cmake.minimum-version = "3.26.4"
```
vyasr, line 2: Would recommend also specifying a couple extra things:
```suggestion
cmake.minimum-version = "3.26.4"
ninja.make-fallback = true
build-dir = "build/{wheel_tag}"
```
Specifying the build-dir is necessary for editable installs, which Graham wants to make sure are supported.

pyproject.toml
```diff
@@ -0,0 +1,10 @@
...
+
+[build-system]
+requires = ["scikit-build-core"]
+build-backend = "scikit_build_core.build"
+
+[project]
+name = "nvjitlink"
+version = "0.1.0"
```
vyasr, line 10: Should we fill in other information here? Things like authors etc? The full list of supported fields is documented [here](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#declaring-project-metadata), and you could look at some RAPIDS packages for examples.

vyasr (changes requested):
A few more things left, but we're very close. Should be able to merge soon after.

CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+Python_add_library(_nvjitlinklib MODULE nvjitlink/_nvjitlinklib.cpp WITH_SOABI)
+
+find_package(
+  CUDAToolkit 12.0 REQUIRED
+)
+target_link_libraries(_nvjitlinklib PRIVATE CUDA::nvJitLink_static CUDA::nvptxcompiler_static)
+
+set(opt_option -O0)
+
```
vyasr, line 19:
```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,20 @@
+[tool.scikit-build]
+cmake.minimum-version = "3.26.4"
+ninja.make-fallback = true
+build-dir = "build/{wheel_tag}"
+sdist.exclude = [".gitignore"] # scikit-build-core fails to create editable installs when hidden files are present
```
vyasr, line 5:
```suggestion
# Workaround for
# https://github.com/scikit-build/scikit-build-core/issues/437
sdist.exclude = [".gitignore"]
```

vyasr (approved):

PR #3, opened 2023-07-19
Skip tests that can corrupt memory

brandon-b-miller:
This PR skips all tests that check error cases, which as of now cause memory corruption and segfaults randomly in the downstream test suite.

### DIFF
--- a/nvjitlink/tests/test_nvjitlink.py
+++ b/nvjitlink/tests/test_nvjitlink.py
@@ -48,7 +48,7 @@ def device_functions_ptx():
 def undefined_extern_cubin():
     return read_test_file('undefined_extern.cubin')
 
-
+@pytest.mark.skip
 def test_create_no_arch_error():
     # nvjitlink expects at least the architecture to be specified.
     with pytest.raises(RuntimeError,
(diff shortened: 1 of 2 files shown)


wence-:
This seems very wrong. If making `nvjitlink` calls can corrupt memory that is either a bug in the C library (and we should report those bugs), or in the wrapper (and we should fix those).

brandon-b-miller:
It is a bug in the underlying c++ library, which has been reported and there's a fix in the pipeline. cc @gmarkall

wence-:
Do know which ctk version it will be fixed in? So that we can condition the skips to `cuda_version < FIXED_VERSION`

wence- (approved):

brandon-b-miller:
> Do know which ctk version it will be fixed in? So that we can condition the skips to `cuda_version < FIXED_VERSION`

Closing the loop on this conversation. Right now we are not sure exactly when this will be fixed. @wence- and I connected offline and agreed this is sufficient for now. https://github.com/rapidsai/nvjitlink/issues/4 tracks this problem.

PR #5, opened 2023-07-20
nvjitlink->pynvjitlink

brandon-b-miller:
This PR renames the library to `pynvjitlink` to avoid confusion with the c++ library.


bdice (approved):

.github/CODEOWNERS
vyasr, line 1: The nvjitlink* teams don't actually exist (and of course neither do the pynvjitlink ones). Can we change all of them in this file to just pynvjitlink-write for now? We can decide at a later date if we really need separate teams for the different components (the package is minimal enough that I don't know if it will be necessary).

I've requested that ops rename the teams accordingly.

vyasr (changes requested):
One small change to make.

.github/CODEOWNERS
brandon-b-miller: Done, I left the `.github` folder under ops.

vyasr (approved):

PR #7, opened 2023-10-16
Support Numba 0.58

gmarkall:
The `NvrtcProgram` class moved to a new module, `numba.cuda.cudadrv.nvrtc` in Numba 0.58 (in PR numba/numba#9086), so we need to import it from there for that version onwards.


pentschev (approved):
LGTM, thanks @gmarkall !

PR #10, opened 2023-10-31
Add CI jobs

brandon-b-miller:
This PR adds a CI workflow that runs on each PR.


pentschev:
I think this has to be under `.github/workflows/pr.yaml`. Additionally the way CI is currently setup probably means you need other steps, for example a `.github/workflows/build.yaml`. Is the intent to make `pynvjitlink` a full conda-forge/wheels package?

brandon-b-miller:
> I think this has to be under `.github/workflows/pr.yaml`. Additionally the way CI is currently setup probably means you need other steps, for example a `.github/workflows/build.yaml`. Is the intent to make `pynvjitlink` a full conda-forge/wheels package?

Yes, we should ultimately have both wheels and conda packages.

pentschev:
Ok, then we probably need to do the entire setup here. That means we will need to copy https://github.com/rapidsai/cudf/blob/branch-23.12/.github/copy-pr-bot.yaml and https://github.com/rapidsai/cudf/blob/branch-23.12/.github/ops-bot.yaml into this PR. We will then need to add conda recipes and wheels build scripts. I can probably help get started here but we will also need help from @bdice and @vyasr who are the experts in packaging.

For the conda recipe, the closest recipe to `pynvjitlink` I know of is perhaps UCX-Py, so what I would do is first copy its [conda files](https://github.com/rapidsai/ucx-py/tree/branch-0.35/conda/recipes/ucx-py) and remove UCX dependencies, probably replace UCX with Numba. We will then need to add [build and tests CI scripts](https://github.com/rapidsai/ucx-py/tree/branch-0.35/ci), which I would do the same, copy from UCX-Py and replace relevant `ucx-py` with `pynvjitlink`. Wheels build and tests scripts are also in the same `ci/` subdirectory and probably can get us a starting point, there are no recipes like in conda but we often require static linking, where @vyasr is definitely someone who we will need help from.

If you want, please ping me on Slack and I can try helping get moving here a bit more, but most definitely we will need more help from Bradley and Vyas later anyway.

brandon-b-miller:
@pentschev we've met once or twice and made some progress planning this out - will loop you in on slack.

ci/build_wheel.sh
```diff
@@ -0,0 +1,28 @@
...
+
+rapids-logger "Building tests"
+
+cd test_binary_generation && make
+
+rapids-logger "Running Tests"
+
+cd ..
+conda install -y pytest
+py.test pynvjitlink/tests
```
bdice, line 28: I think I would prefer to run tests in a separate job that depends on the build job.

ci/build_wheel.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# install CUDAToolKit version 12.2
+apt-get update
+apt-get -y install cuda-toolkit-12-2
+
+rapids-logger "Check conda environment"
+
+conda list
+
+rapids-logger "Building wheel"
```
bdice, line 12: Change tense.
```suggestion
rapids-logger "Build wheel"
```

ci/build_wheel.sh
```diff
@@ -0,0 +1,28 @@
...
+
+pip wheel .
+
+rapids-logger "Installing wheel"
```
bdice, line 16:
```suggestion
rapids-logger "Install wheel"
```

ci/build_wheel.sh
```diff
@@ -0,0 +1,28 @@
...
+
+pip install pynvjitlink-0.1.0-cp310-cp310-manylinux_2_35_x86_64.whl
+
+rapids-logger "Building tests"
```
bdice, line 20:
```suggestion
rapids-logger "Build tests"
```

ci/build_wheel.sh
```diff
@@ -0,0 +1,28 @@
...
+
+cd test_binary_generation && make
+
+rapids-logger "Running Tests"
```
bdice, line 24:
```suggestion
rapids-logger "Run tests"
```

ci/check_style.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+# Copyright (c) 2023, NVIDIA CORPORATION
```
bdice, line 2: Borrowing from https://github.com/rapidsai/rapids-cmake/blob/branch-23.12/ci/check_style.sh, but simplified. We could add dependencies.yaml and use the same script as the rest of RAPIDS if desired.
```suggestion
# Copyright (c) 2023, NVIDIA CORPORATION

set -euo pipefail

pip install pre-commit

# Run pre-commit checks
pre-commit run --all-files --show-diff-on-failure
```

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,37 @@
...
+    uses: rapidsai/shared-action-workflows/.github/workflows/pr-builder.yaml@branch-23.10
+  checks:
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/checks.yaml@branch-23.10
+    with:
+      enable_check_generated_files: false
+  wheel-build:
+    runs-on: ubuntu-latest
+    container:
+      image: "rapidsai/ci-conda:latest"
```
bdice, line 27:
```suggestion
      # Use CentOS 7 for an old glibc version, so the wheel has broad manylinux compatibility.
      image: "rapidsai/ci-conda:cuda12.0.1-centos7-py3.10"
```

ci/build_wheel.sh
```diff
@@ -0,0 +1,28 @@
...
+
+conda list
+
+rapids-logger "Building wheel"
+
+pip wheel .
+
+rapids-logger "Installing wheel"
+
+pip install pynvjitlink-0.1.0-cp310-cp310-manylinux_2_35_x86_64.whl
```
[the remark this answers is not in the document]
bdice: I just noticed that we're going to need a Python version matrix. This wheel only support Python 3.10. We need to ship pynvjitlink on Python 3.9, 3.10 for RAPIDS, and also 3.11 and 3.12 if we want to make external consumers from Numba happy. We don't have Python 3.11 / 3.12 Docker images yet.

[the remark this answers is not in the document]
bdice: @gmarkall Can you comment on the expected Python version support for this package?

ci/build_wheel.sh
brandon-b-miller: Roger that, I was wondering the same thing. Do we think it's worth doing both in this one PR or would it be better to break off the testing phase into its own PR to avoid muddying things?

brandon-b-miller:
@bdice do you know why the ctk installation fails on centos by any chance? details below.

<details>

```
--> Finished Dependency Resolution
Error: Package: nsight-systems-2023.3.3-2023.3.3.42_233333266658v0-0.x86_64 (cuda)
           Requires: libOpenGL.so.0()(64bit)
           Available: 1:libglvnd-opengl-1.0.1-0.6.git5baa1e5.el7.x86_64 (cuda)
               libOpenGL.so.0()(64bit)
           Available: 1:libglvnd-opengl-1.0.1-0.8.git5baa1e5.el7.x86_64 (base)
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-367.48-1.el7.x86_64 (cuda)
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-375.26-1.el7.x86_64 (cuda)
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-375.51-1.el7.x86_64 (cuda)
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-375.74-1.el7.x86_64 (cuda)
 You could try using --skip-broken to work around the problem
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-375.88-1.el7.x86_64 (cuda)
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-384.66-1.el7.x86_64 (cuda)
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-384.81-1.el7.x86_64 (cuda)
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-384.111-1.el7.x86_64 (cuda)
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-384.145-1.el7.x86_64 (cuda)
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-384.183-1.el7.x86_64 (cuda)
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-387.26-1.el7.x86_64 (cuda)
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-390.12-1.el7.x86_64 (cuda)
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-390.30-1.el7.x86_64 (cuda)
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-[390](https://github.com/rapidsai/pynvjitlink/actions/runs/6734331671/job/18305010392?pr=10#step:5:391).116-1.el7.x86_64 (cuda)
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-396.26-1.el7.x86_64 (cuda)
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-396.37-1.el7.x86_64 (cuda)
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-[396](https://github.com/rapidsai/pynvjitlink/actions/runs/6734331671/job/18305010392?pr=10#step:5:397).44-1.el7.x86_64 (cuda)
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-396.82-1.el7.x86_64 (cuda)
               libOpenGL.so.0()(64bit)
           Available: 1:xorg-x11-drv-nvidia-gl-346.46-1.el7.x86_64 (cuda)
               Not found
           Available: 1:xorg-x11-drv-nvidia-gl-352.39-1.el7.x86_64 (cuda)
               Not found
           Available: 1:xorg-x11-drv-nvidia-gl-352.55-1.el7.x86_64 (cuda)
               Not found
           Available: 1:xorg-x11-drv-nvidia-gl-352.63-1.el7.x86_64 (cuda)
               Not found
           Available: 1:xorg-x11-drv-nvidia-gl-352.68-1.el7.x86_64 (cuda)
               Not found
           Available: 1:xorg-x11-drv-nvidia-gl-352.79-1.el7.x86_64 (cuda)
               Not found
           Available: 1:xorg-x11-drv-nvidia-gl-352.93-1.el7.x86_64 (cuda)
               Not found
           Available: 1:xorg-x11-drv-nvidia-gl-352.99-1.el7.x86_64 (cuda)
               Not found
 You could try running: rpm -Va --nofiles --nodigest
```
</details>

ci/build_wheel.sh
bdice: Let's do it in this PR. We could use a wider test matrix than our build matrix.

bdice:
> @bdice do you know why the ctk installation fails on centos by any chance? details below.

Not sure. Is there a package with fewer dependencies? We don't need NSight for this package to build.

ci/build_wheel.sh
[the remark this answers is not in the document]
gmarkall: I was hoping to follow [SPEC 0](https://scientific-python.org/specs/spec-0000/#support-window), so at the present time 3.10-3.12 support would be my aim. If 3.9 support is also needed for RAPIDS then that seems OK. If we can't initially provide wheels for 3.11 and 3.12 then I think that's also OK.

My reasoning is that for non-RAPIDS use cases, it will always be possible for users to build pynvjitlink themselves (unlike with cubinlinker), and we're covering the RAPIDS deployment scenarios in supporting 3.9 and 3.10 from the start. If there's no short-term route to subsequently adding 3.11 and 3.12, I'd like to look into how to get to supporting them eventually.

gmarkall:
I think the build is presently failing because the default standard is C++03 in GCC 4.8.5, and there isn't an explicit `-std=c++11` flag. Hopefully C++11 is sufficient, as it's the last C++ standard supported by GCC 4.8.

ci/build_wheel.sh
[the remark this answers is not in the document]
brandon-b-miller: It looks like there aren't `ci-conda` centos images spanning the range of python versions we need. Should we be installing the right version of python ourselves within the container or is there something else I should be doing instead?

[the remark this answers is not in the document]
bdice: You can add newer Python CI images, or just use Python versions from conda. Perhaps it will suffice to have a single build job that iterates over all the Python versions, since the builds should be fast.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,42 @@
...
+  checks:
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/checks.yaml@branch-23.12
+    with:
+      enable_check_generated_files: false
+  wheel-build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.10, 3.11, 3.12]
```
bdice, line 29: FYI, you have to wrap 3.10 in quotes because it's currently being interpreted as a float 3.1.
```suggestion
        python-version: ["3.10", "3.11", "3.12"]
```

ci/build_wheel.sh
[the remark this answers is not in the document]
bdice: > What do you mean by "add newer python CI images"?

Today, our CI images only cover Python 3.9 and 3.10. https://github.com/rapidsai/ci-imgs/blob/bf7d540855e7e94695a54e40080c3acb8681764a/matrix.yaml#L8-L10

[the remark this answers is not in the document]
bdice: > Perhaps it will suffice to have a single build job that iterates over all the Python versions, since the builds should be fast.

I had in mind using a conda variant build like in ucxx, where the `conda_build_config.yaml` has a list of Python versions and `conda build` knows to build all of them.

https://github.com/rapidsai/ucxx/blob/4464ca942d5ac71aabe4492e2a6d59da453f06dc/conda/recipes/ucxx/conda_build_config.yaml#L19-L21

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,70 @@
...
+      with:
+        fetch-depth: 0
+    - name: Get PR Info
+      id: get-pr-info
+      uses: rapidsai/shared-actions/get-pr-info@main
+    - name: Download Python Wheel
+      uses: actions/download-artifact@v2
+      with:
+        path: ./
+    - name: run test_wheel.sh
```
bdice, line 69:
```suggestion
    - name: Run test_wheel.sh
```

brandon-b-miller:
@bdice is there a container that would be suitable for testing that does _not_ have cuda built in at all? It isn't needed at all on the testing machines because we statically link the two pieces of it that we need in the build step and it's actually getting in the way of testing due to https://nvbugswb.nvidia.com/NvBugs5/SWBug.aspx?bugid=4375843&cmtNo=.

bdice:
(Unless you're not relying on gha-tools or other RAPIDS CI tooling and can use a plain base container like `ubuntu`.)

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,77 @@
...
+      enable_check_generated_files: false
+  wheel-build:
+    runs-on: linux-${{ matrix.arch }}-cpu4
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.10"]
+        arch: ["arm64", "amd64"]
+    container:
+      # Use CentOS 7 for an old glibc version, so the wheel has broad manylinux compatibility.
+      image: "rapidsai/ci-wheel:cuda12.0.1-centos7-py${{ matrix.python-version }}"
```
brandon-b-miller, line 34:
```suggestion
      image: "rapidsai/ci-wheel:cuda12.0.1-rockylinux8-py${{ matrix.python-version }}"
```

pynvjitlink/tests/test_pynvjitlink_api.py
```diff
@@ -98,13 +104,15 @@ def test_add_cubin_with_fatbin_error(device_functions_fatbin):
         nvjitlinker.add_cubin(device_functions_fatbin, name)
 
 
+@pytest.mark.skip
```
brandon-b-miller, line 107: TODO: not sure why we need to skip this in CI but it works for me locally.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,103 @@
...
+      enable_check_generated_files: false
+  wheel-build:
+    runs-on: linux-${{ matrix.arch }}-cpu4
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.10"]
+        arch: ["arm64", "amd64"]
+    container:
+      # Use CentOS 7 for an old glibc version, so the wheel has broad manylinux compatibility.
+      image: "rapidsai/ci-wheel:cuda12.0.1-rockylinux8-py${{ matrix.python-version }}"
```
bdice, line 35: The comment says CentOS 7 but the image is Rocky 8. How do we reconcile this? I think we want to build the amd64 wheels on CentOS 7 and arm64 on Rocky 8, to align with our preferred RAPIDS manylinux tags.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,103 @@
...
+  wheel-test-amd64:
+    needs:
+      - wheel-build
+    runs-on: linux-amd64-gpu-v100-latest-1
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.10"]
+        linux-version: ["centos7", "rockylinux8"]
+    container:
+      image: "rapidsai/citestwheel:cuda12.0.1-ubuntu20.04-py${{ matrix.python-version }}"
```
bdice, line 61: Is `linux-version: ["centos7", "rockylinux8"]` being used? I would've expected it here.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,103 @@
...
+    - name: run test_wheel.sh
+      run: ci/test_wheel.sh 
+  wheel-test-arm64:
+    needs:
+      - wheel-build
+    runs-on: linux-arm64-gpu-a100-latest-1
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.10"]
+        linux-version: ["centos7", "rockylinux8"]
```
bdice, line 85: Same question as above about `linux-version` being unused.

CMakeLists.txt
```diff
@@ -1,17 +1,19 @@
+set(CMAKE_CXX_STANDARD 11)
```
bdice, line 1: Put this after `cmake_minimum_required`. And also consider using `target_compile_features(pynvjitlink PRIVATE cxx_std_11)` instead, if my understanding of this conversation is correct: https://github.com/rapidsai/cudf/pull/13795#discussion_r1393154975

ci/build_wheel.sh
```diff
@@ -0,0 +1,23 @@
...
+yum update
+yum install -y epel-release
+yum install -y nvidia-driver-latest-dkms
+yum install -y cuda-toolkit-12-3
+
+rapids-logger "Install GCC"
+yum install -y centos-release-scl
+yum install -y devtoolset-9
+
+source scl_source enable devtoolset-9
```
bdice, line 15: Can you give some context on this? We're using `devtoolset-11` in https://github.com/rapidsai/ci-imgs/blob/main/ci-wheel.Dockerfile

ci/build_wheel.sh
```diff
@@ -0,0 +1,23 @@
...
+
+rapids-logger "Check conda environment"
+conda list
+
+rapids-logger "Build wheel"
+export SCCACHE_S3_NO_CREDENTIALS=1
```
bdice, line 21: Why is this needed? I think we're probably missing some other piece of setup like https://github.com/rapidsai/shared-workflows/blob/9638706c8c3e43bb0b210528a814a2abb2f3a923/.github/workflows/wheels-build.yaml#L126-L130

ci/test_wheel.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+# Copyright (c) 2023, NVIDIA CORPORATION
+
+set -e
+
+rapids-logger "Search artficact directory"
+ls ./artifact
+
```
bdice, line 8:
```suggestion
rapids-logger "Search artifact directory"
ls ./artifact
```

ci/test_wheel.sh
```diff
@@ -0,0 +1,19 @@
...
+
+rapids-logger "Install wheel"
+pip install  --find-links ./artifact pynvjitlink
```
bdice, line 11:
```suggestion
pip install --find-links ./artifact pynvjitlink
```

ci/test_wheel.sh
```diff
@@ -0,0 +1,19 @@
...
+
+rapids-logger "Build Tests"
+cd test_binary_generation && make
+
+rapids-logger "Run Tests"
+cd ..
+pip install pytest
+py.test pynvjitlink/tests
```
bdice, line 19:
```suggestion
pytest pynvjitlink/tests
```

pynvjitlink/tests/test_pynvjitlink.py
```diff
@@ -188,7 +192,7 @@ def test_get_linked_ptx_from_lto(device_functions_ltoir):
     handle = _nvjitlinklib.create('-arch=sm_75', '-lto', '-ptx')
     _nvjitlinklib.add_data(handle, input_type, data, filename)
     _nvjitlinklib.complete(handle)
-    ptx = _nvjitlinklib.get_linked_ptx(handle)
+    ptx = _nvjitlinklib.get_linked_ptx(handle)  # noqa: F841
```
bdice, line 195: Why is this the right fix, rather than removing the unused local variable?
```suggestion
    _nvjitlinklib.get_linked_ptx(handle)
```

pynvjitlink/tests/test_pynvjitlink_api.py
bdice: Maybe this should be xfailed in CI instead of skipped?

.github/workflows/pr.yaml
brandon-b-miller: I originally switched this because on the arm runners, there appeared to be no matching container
```
no matching manifest for linux/arm64/v8 in the manifest list entries
```
whereas there were rockylinux containers available for arm and x86. However I think I can implement what you are saying by separating the build phase into two pieces like the testing phase is. I'll try and update here.

.github/workflows/pr.yaml
brandon-b-miller: This has been updated. To be honest I'm not 100% sure what should be here. I noticed that in cuDF we only test wheels on ubuntu, albeit testing on two versions, and only testing arm with ubuntu 20.04. This seems to line up with what's available for docker under `rapidsai/citestwheel`, so I switched it to mirror that here, let me know if that seems to line up with what we want.

pynvjitlink/tests/test_pynvjitlink_api.py
brandon-b-miller: I think we still want to skip things due to https://github.com/rapidsai/pynvjitlink/pull/3

.github/workflows/pr.yaml
brandon-b-miller: this has been done.

ci/build_wheel.sh
brandon-b-miller: This was needed when we used previous images that didn't have a recent enough gcc, and has been removed.

ci/build_wheel.sh
brandon-b-miller: Without this, I get 
```
      sccache /usr/bin/c++    -o CMakeFiles/cmTC_1b1f2.dir/testCXXCompiler.cxx.o -c /__w/pynvjitlink/pynvjitlink/build/cp310-cp310-manylinux_2_17_x86_64/CMakeFiles/CMakeScratch/TryCompile-DBkCuI/testCXXCompiler.cxx
      sccache: error: Timed out waiting for server startup. Maybe the remote service is unreachable?
```
during the wheel build step, this is a workaround. Still following up on this.

quasiben:
@gmarkall and I met to fix up tests.  I'll follow up in a separate PR to fix those tests.  For now it seems like PR is ready for review then merge

ci/build_wheel.sh
bdice: I think the follow-up task here is to try using the `aws-actions/configure-aws-credentials@v3` step that I linked above. It doesn't matter for this PR, though.

bdice (approved):
This looks good to me. I think we can merge this and test the artifacts, if you agree?

bdice:
This adds CI to PRs -- but we will need a `build.yaml` to make this publish the wheels to a nightly index for testing. @brandon-b-miller and @quasiben, are we ready for that? I think so?

gmarkall (approved):
I think this is good to merge - there are some follow-ups that I think we need like test amendments (and figuring out why the library is linked against `libcuda.so`), but it would be good to get this in then address those in follow-up PRs.

brandon-b-miller:
> This adds CI to PRs -- but we will need a `build.yaml` to make this publish the wheels to a nightly index for testing (maybe triggered on tags?). @brandon-b-miller and @quasiben, are we ready for that? I think so?

This seems like a logical next step to me

PR #11, opened 2023-11-02
Add support for LTO in Numba linker

gmarkall:
Support for enabling LTO via nvjitlink is added - it is enabled by passing the `lto=True` kwarg. I also made a couple of other related changes:

* The minimum CUDA toolkit version is bumped to 12.3 - there's no point building against anything older, especially for LTO use cases.
* I noticed there were no tests of the patch, so I ported those from CubinLinker. Note that although CubinLinker is not open-source, the source of its tests is included in the distribution, so this addition only brings things that were already public into pynvjitlink.

cc @leopoldcambier to ensure this meets your requirements for LTO use cases.


wence- (approved):
Looks reasonable to me, I think.

vyasr (approved):
Seems fine especially since it's largely just forwarding along arguments.

gmarkall:
I just merged `main` so that CI would run on this PR now that #10 is merged.

Issue #4, opened 2023-07-20
[BUG] Remove xfails on error checking test cases when possible


brandon-b-miller:
PR https://github.com/rapidsai/nvjitlink/pull/3 skips some tests that corrupt memory due to an underlying c++ bug. We should remove these when a future CUDA version is released that fixes said bug in `libnvjitlink`.

gmarkall:
We no longer have any xfails in the test suite.

PR #12, opened 2023-11-17
Additional cleanup

quasiben:
PR cleans up a few remaining items before release:
- Metdata in pyproject.toml
- Remove skips and update tests


brandon-b-miller:
Does this close https://github.com/rapidsai/pynvjitlink/issues/4 ?

brandon-b-miller (approved):

pynvjitlink/tests/test_pynvjitlink.py
```diff
@@ -49,30 +49,26 @@ def undefined_extern_cubin():
     return read_test_file('undefined_extern.cubin')
 
 
-@pytest.mark.skip
```
wence-, line 52: I think this was fixed in CTK 12.2, but we're not installing environments with that yet, so I am not sure if we can skip these?

brandon-b-miller: We require cuda 12.3 to build the project now https://github.com/rapidsai/pynvjitlink/blob/main/CMakeLists.txt#L14

wence-: Ah, OK. Is that also the version that the end user gets?

brandon-b-miller: Yes - we build in the two pieces of the CTK that are used statically so the user should end up with the 12.3 ctk binaries on their system regardless of what they have installed.

gmarkall (approved):
Changes look good to me.

wence- (approved):

PR #13, opened 2023-11-20
Remove compiled files

brandon-b-miller:
This PR removes some compiled files which are both not useful and should be getting rebuilt in every test environment anyways.


bdice (approved):
Can we add .gitignore entries for built files?

PR #15, opened 2023-11-20
Add a workflow which builds wheels on tag creation

brandon-b-miller:
This PR adds a second github actions workflow which runs when a tag is created. Ideally this works as a release pipeline.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,120 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.
+name: release
+
+on:
+  create:
+    tags:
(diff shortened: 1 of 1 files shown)


wence-:
Do you expect that the contents will change relative to the `pr.yaml` case? Otherwise you can just select `on` for both PRs and tag creation in the one file.

brandon-b-miller:
> Do you expect that the contents will change relative to the `pr.yaml` case? Otherwise you can just select `on` for both PRs and tag creation in the one file.

Yeah - at some this workflow will be responsible for uploading the resulting artifacts to pypi for instance, whereas the PR workflow won't.

PR #16, opened 2023-11-22
Clean up style checks.

bdice:
This PR updates style checks to trim trailing whitespace and run a basic suite of Python linting/formatting with ruff.


brandon-b-miller (approved):
Thanks @bdice !

PR #17, opened 2023-11-22
Create releases when tags are pushed.

bdice:
This PR updates the CI workflows to create a nightly wheel when a tag is pushed. This is loosely what was discussed in #15. @brandon-b-miller Feel free to adapt this or close it.


.github/workflows/pr.yaml
```diff
@@ -1,13 +1,15 @@
...
   push:
     branches:
       - "pull-request/[0-9]+"
+    tags:
+      - 'v*'
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
   cancel-in-progress: true
```
wence-, line 13: In theory this would mean if two tags are pushed concurrently, only the later running job would end up uploading a wheel. Probably this is unlikely to happen because we don't push tags very often (I think?)

I am also not sure that the `event_name` part disambiguates anything, since in both a branch-based push or a tag-based push the `event_name` is just `"push"`. Am I missing something?

.github/workflows/pr.yaml
```diff
@@ -118,3 +121,17 @@ jobs:
         path: ./
     - name: run test_wheel.sh
       run: ci/test_wheel.sh
+  wheel-publish-pynvjitlink:
+    needs:
+      - wheel-test-amd64
+      - wheel-test-arm64
+    # Only trigger this step for tags
+    if: startsWith(github.event_name, 'push') && github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
```
wence-, line 129:
```suggestion
    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
```

`startsWith(foo, "push")` is unnecessary if also checking for equality with `"push"`.

.github/workflows/pr.yaml
```diff
@@ -18,6 +20,7 @@ jobs:
       - wheel-build-arm64
       - wheel-test-amd64
       - wheel-test-arm64
+      - wheel-publish-pynvjitlink
```
wence-, line 23: This applies a dependency to a job that is skipped unless a tag is pushed, and so will only succeed on tag pushes, I think.

brandon-b-miller: this maybe resolved by https://github.com/rapidsai/pynvjitlink/pull/17/commits/a854b453eee6989336a139ac538b9e56510b72bd

.github/workflows/pr.yaml
bdice: @wence- Do you have a fix for these issues in mind?

.github/workflows/pr.yaml
brandon-b-miller: This appears to be resolved if we're ok with the above.

.github/workflows/pr.yaml
brandon-b-miller: The second issue might not be a huge concern yet either if the only drawback is that release jobs could compete for resources with PR jobs IMO.

brandon-b-miller (approved):

PR #18, opened 2023-11-27
Disambiguate cuda major versions in wheel names

brandon-b-miller:
This PR labels the wheel name with a `-cu12` suffix.

### DIFF
--- a/ci/test_wheel.sh
+++ b/ci/test_wheel.sh
@@ -7,7 +7,7 @@ rapids-logger "Search artifact directory"
 ls ./artifact
 
 rapids-logger "Install wheel"
-pip install --find-links ./artifact pynvjitlink
+pip install --find-links ./artifact pynvjitlink-cu12
 
 rapids-logger "Build Tests"
 cd test_binary_generation && make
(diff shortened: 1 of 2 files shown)


brandon-b-miller:
Should this change affect the actual name of the package within python?
```
ImportError: cannot import name 'patch' from 'pynvjitlink' (unknown location)
```

bdice:
Maybe you need something to specify the package location since it doesn't match the name:

```
[tool.setuptools]
packages = { "find" = { where = ["pynvjitlink"] } }
```

(Borrowed from https://github.com/rapidsai/dependency-file-generator/blob/78ca9ecbf01edaff93f898f2202fb34342f3820e/pyproject.toml#L31)

bdice:
This should help: https://scikit-build-core.readthedocs.io/en/latest/configuration.html#customizing-the-built-wheel

> The wheel will automatically look for Python packages at src/<package_name>, python/<package_name>, and <package_name>, in that order. If you want to list packages explicitly, you can. The final path element is the package.
>
> ```
> [tool.scikit-build]
> wheel.packages = ["python/src/mypackage"]
> ```

Try this:
```
[tool.scikit-build]
wheel.packages = ["pynvjitlink"]
```

PR #20, opened 2023-11-28
Separate pull request and tag workflows

brandon-b-miller:
This PR implements a two-workflow approach where one workflow is responsible for pull requests and another is responsible for tags and uploading wheels.

### DIFF
--- a/.github/workflows/build_and_test_wheels.yaml
+++ b/.github/workflows/build_and_test_wheels.yaml
@@ -1,0 +1,91 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.
+jobs:
+  build-and-test-wheels:
+    needs:
+      - wheel-build-amd64
+      - wheel-build-arm64
+      - wheel-test-amd64
+      - wheel-test-arm64
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/pr-builder.yaml@branch-23.12
+  wheel-build-amd64:
+    runs-on: linux-amd64-cpu4
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.10"]
+    container:
+      image: "rapidsai/ci-wheel:cuda12.0.1-centos7-py${{ matrix.python-version }}"
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: Run build_wheel.sh
+      run: ci/build_wheel.sh
+    - name: Upload Python Wheel
+      uses: actions/upload-artifact@v3
+      with:
+        path: ./wheel-build
+  wheel-build-arm64:
+    runs-on: linux-arm64-cpu4
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.10"]
+    container:
+      image: "rapidsai/ci-wheel:cuda12.0.1-rockylinux8-py${{ matrix.python-version }}"
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: Run build_wheel.sh
+      run: ci/build_wheel.sh
+    - name: Upload Python Wheel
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -0,0 +1,38 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}-release
+  cancel-in-progress: true
+
+jobs:
+  pr-builder:
+    needs:
+      - checks
+      - build-and-test-wheels
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/pr-builder.yaml@branch-23.12
```
bdice, line 19: This repo was renamed to https://github.com/rapidsai/shared-workflows. Please update `shared-action-workflows` to `shared-workflows` accordingly.

bdice: Also let's use `branch-24.02` since 23.12 is in freeze. We can go ahead and target this as if it's being released with 24.02.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,38 @@
...
+  checks:
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/checks.yaml@branch-23.12
+    with:
+      enable_check_generated_files: false
+      enable_check_pr_job_dependencies: false
```
bdice, line 25: Do we still need to disable these? Or was that only necessary with the previous approach?

.github/workflows/build_and_test_wheels.yaml
```diff
@@ -0,0 +1,94 @@
...
+      - wheel-build-arm64
+      - wheel-test-amd64
+      - wheel-test-arm64
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/pr-builder.yaml@branch-23.12
+  wheel-build-amd64:
+    runs-on: linux-amd64-cpu4
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.10"]
```
bdice, line 18: We'll want to build for Python 3.9-3.12 at some point. We may need new `ci-wheel` images for that.

.github/workflows/build_and_test_wheels.yaml
```diff
@@ -0,0 +1,94 @@
...
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: Download Python Wheel
+      uses: actions/download-artifact@v3
+      with:
+        path: ./
+    - name: run test_wheel.sh
```
bdice, line 93: Use title case (or some consistent casing that isn't all lowercase) in all the step names.
```suggestion
    - name: Run test_wheel.sh
```

bdice (approved):
This looks pretty close to what we want. I have a few nits and some future work outlined. Let's file issues for anything that looks like future work.

.github/workflows/build.yaml
brandon-b-miller: updated

.github/workflows/build.yaml
brandon-b-miller: I think the second one isn't necessary, I removed it for now.

.github/workflows/build_and_test_wheels.yaml
brandon-b-miller: right - we did a bunch of playing around with which containers to use early on and ended up rolling with the current matrix mostly _because_ that's what the existing images support on the rapids side. Ultimately however `pynvjitlink` has consumers outside of rapids that do require other versions of python as you mention, so we'll eventually either need to move away from rapids images or create rapids images beyond what rapids supports more broadly just for building this package. Not sure which would be better.

bdice: I opened https://github.com/rapidsai/miniforge-cuda/pull/55 to help start this process.

AyodeAwe (approved):

PR #23, opened 2023-11-28
Fix release workflow

brandon-b-miller:


.github/workflows/build.yaml
```diff
@@ -13,15 +13,9 @@ concurrency:
 jobs:
   pr-builder:
     needs:
-      - checks
       - build-and-test-wheels
     secrets: inherit
     uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@branch-24.02
```
vyasr, line 18: You can get rid of this altogether.

ci/build_wheel.sh
```diff
@@ -15,3 +15,6 @@ rapids-logger "Build wheel"
 export SCCACHE_S3_NO_CREDENTIALS=1
 mkdir -p ./wheel-build
 pip wheel . --wheel-dir=./wheel-build -vvv
+
+rapids-logger "Upload Wheel"
+RAPIDS_BUILD_TYPE="branch" rapids-upload-to-s3 pynvjitlink-cu12 ./wheel-build/
```
vyasr, line 20: You'll probably need to add corresponding download commands to the test_wheel.sh script.

.github/workflows/pr.yaml
```diff
@@ -23,5 +23,31 @@ jobs:
...
+          export MATRIX="
+          - { CUDA_VER: '12.0.1', ARCH: 'amd64', PY_VER: '3.9', LINUX_VER: 'centos7' }
+          - { CUDA_VER: '12.0.1', ARCH: 'amd64', PY_VER: '3.10', LINUX_VER: 'centos7' }
+          - { CUDA_VER: '12.0.1', ARCH: 'arm64', PY_VER: '3.9', LINUX_VER: 'centos7' }
+          - { CUDA_VER: '12.0.1', ARCH: 'arm64', PY_VER: '3.10', LINUX_VER: 'centos7' }
+          "
+
+          echo "MATRIX=$(
+            yq -n -o json 'env(MATRIX)' | \
+            jq -c '${{ inputs.matrix_filter }} | {include: .}' \
```
ajschmidt8, line 45:
```suggestion
            jq -c '.' \
```

This will need to be tweaked so that the JSON is in the proper format for the shared-workflow to consume.

.github/workflows/pr.yaml
```diff
@@ -23,5 +24,62 @@ jobs:
...
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@branch-24.02
+    with:
+      build_type: pull-request
+      script: "ci/build_wheel.sh"
+      matrix_filter: ${{ needs.compute-matrix-build.outputs.MATRIX }}
+  test-wheels:
+    needs:
+      - build-wheels
+      - compute-matrix-test
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-test.yaml@branch-24.02
```
brandon-b-miller, line 81: TODO: reuse `wheels-build` to test on CPU machines

.github/workflows/build_and_test_wheels.yaml
vyasr, line 1: I don't think we want this file to exist anymore. In other repos we just duplicate the bits that need duplication between build.yaml and pr.yaml, and there are chunks of this that really only make sense for the pr.yaml workflow anyway.

ajschmidt8: We could dedupe the matrix by creating a composite action in this repository. I can share more on that tomorrow when I'm back in front of my desktop.

.github/workflows/pr.yaml
ajschmidt8: This resulted in the errors below, so we reverted to testing on GPU machines:

- https://github.com/rapidsai/pynvjitlink/actions/runs/7049358912/job/19187956792#step:9:110

We can revisit this in the future.

Marking as resolved for now.

ajschmidt8 (approved):

brandon-b-miller:
@vyasr any last thoughts here before we merge?

vyasr (approved):

PR #33, opened 2024-01-05
Add Conda packages and CI workflows

brandon-b-miller:
This PR adds build scripts for creating and testing conda packages and the associated CI workflows.


ci/build_conda.sh
```diff
@@ -0,0 +1,19 @@
...
+set -euo pipefail
+
+source rapids-env-update
+
+export CMAKE_GENERATOR=Ninja
+
+rapids-print-env
+
+rapids-logger "Install CUDA Toolkit"
+source "$(dirname "$0")/install_latest_cuda_toolkit.sh"
```
msarahan, line 13: This process of installing system-wide cudatoolkit feels like an antipattern to me. It goes against what conda is good at, and I'm concerned that it might make things more brittle. As such, the work that we are needing to do now to address confusion with finding libraries feels like wasted effort.

The fundamental issue is that appropriate cudatoolkit libraries are not available on Conda, right? Is there a different approach we could take here, where we work on addressing that fundamental issue first and then use more standard conda patterns in this library?

Have I misunderstood what is going on here? Are there other reasons to be using a system install of the CTK?

jakirkham: If we want to build with CUDA 12.2, those libraries are available

pyproject.toml
```diff
@@ -1,6 +1,8 @@
 [tool.scikit-build]
 cmake.minimum-version = "3.26.4"
 cmake.verbose = true
+cmake.args = ["-DCUDAToolkit_ROOT=/usr/local/cuda-12.3", "--debug-find"]
```
jakirkham, line 4: Though it looks like we are building with CUDA 12.3 here. Do we need CUDA 12.3? Or could we build with CUDA 12.2?

ci/build_conda.sh
brandon-b-miller: The main issue is that the package needs to ship the latest version of `libnvjitlink_static` for it to be able to handle the PTX from the cuDF build. So far we've used the same pattern for the pip packages.

pyproject.toml
brandon-b-miller: The `12.3` version of `libnvjitlink` has a couple of bugfixes we could use. cc @gmarkall for details.

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+    #- libnvjitlink-dev
+    #- libnvjitlink-static
+    #- cuda-version ={{ cuda_version }}
+    - python
+    - pip
+    - scikit-build-core
+  run:
+    - python
+    - numba >=0.57
+    - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
```
jakirkham, line 41: If we did build with CUDA 12.2 and wanted to make sure this didn't get installed with CUDA 12.3, we could add an upper bound to the `cuda-version` installed. Perhaps like this?

```suggestion
    - {{ pin_compatible('cuda-version', min_pin='x', max_pin='12.3') }}
```

ci/build_conda.sh
msarahan: I'm pretty confused here, so please correct me where I have no doubt gotten something wrong.

* We need pynvjitlink in order to finish our CUDA 12.2 release
* We can't use our existing parts of CUDA 12.2 to build pynvjitlink, because we need libnvjitlink from CUDA 12.3

Is that right? If so, I suppose you have no choice but the system packages, but it also seems like a really awkward position to be in. How was pynvjitlink both mandatory to 12.2 and not functional until 12.3?

pyproject.toml
bdice: [nvbug 4375843](https://nvbugswb.nvidia.com/NVBugs5/redir.aspx?url=/4375843) is one issue that I know of but there may be others.

ci/build_conda.sh
bdice: Great questions. I can try to illuminate the goals, design, and requirements, which might help.

nvjitlink is a library added in CUDA 12 that provides "a set of APIs which can be used at runtime to link together GPU device code". pynvjitlink provides Python bindings to that code. cuDF ships a bundle of PTX code generated at build time that is used by Numba and linked together with Numba-generated code. This process needs pynvjitlink in order to provide backward compatibility, if the PTX bundled with cuDF at build time is too new to run on the user's driver.

For example, if we build cuDF with CUDA 12.2, it will generate and bundle some CUDA 12.2 PTX code with the library. Numba wants to link this with user-provided code but the result will not run on driver 525 (which shipped alongside CUDA 12.0) because it doesn't know how to use the newer PTX code. In order for RAPIDS to migrate to building with CUDA 12.x (x > 0), we must depend on pynvjitlink.

The goal is to release a new pynvjitlink (which statically links libnvjitlink) for each new CUDA Toolkit minor version. (Other reasons to make a new version would include other matrix changes like Python version updates or updates to the pynvjitlink bindings themselves.) This release process is _not_ fundamentally synced to the RAPIDS release cycle, and will likely involve intermittent but off-cycle releases that are not aligned to RAPIDS 24.02, 24.04, etc.

We want (need?) to build with the newest possible version of nvjitlink for compatibility reasons. RAPIDS wants to move to building with CUDA 12.2, so it's reasonable to ask why we can't just use nvjitlink from CUDA 12.2 (which has conda-forge packages). There are some additional bugs in nvjitlink that were only recently resolved in 12.3 (see other thread: https://github.com/rapidsai/pynvjitlink/pull/33#discussion_r1445510368), so this isn't viable. We also want `pynvjitlink` to be usable by Numba consumers outside of RAPIDS, so supporting PTX from CUDA "12.latest" (and not just limiting to the CUDA build version used by RAPIDS) is important.

Generally, conda packages have lagged other channel sources (like RPM packages) in availability. Because we _need_ to build pynvjitlink against CUDA "12.latest" for pip users (who could have any driver/toolkit), we chose to pull nvjitlink from RPM packages. It is possible for conda packages of pynvjitlink to take a different strategy, but my initial hope was that we could just do the same thing and pull the latest CUDA RPM packages for the conda build. Clearly there are hiccups with CMake and conda support around this since it didn't work out of the box. I am certainly open to other strategies here but it will be difficult to maintain pynvjitlink if we are pulling the nvjitlink from two different sources with two different versions, because we want to make new releases for both pip and conda simultaneously.

We require the nvjitlink library from CUDA 12.3+ to proceed. CUDA 12.3 is not on `conda-forge` yet but that still wouldn't necessarily solve our problem if conda-forge continues to lag mainstream availability (we need to release pip and conda support together). Perhaps we could pull from `nvidia` channel packages or fix the issues with using system packages (which I trust to be released on "day one" of a new minor version).

conda/recipes/pynvjitlink/meta.yaml
bdice: Good question but I'm not sure if we need constraints at runtime. The issue is needing to build with nvjitlink from CUDA 12.3 due to bugs in earlier versions. https://github.com/rapidsai/pynvjitlink/pull/33#discussion_r1445510368

pyproject.toml
bdice: @brandon-b-miller / @gmarkall Is that the same issue tracked in #3 / #4?

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+build:
+  noarch: python
+  script:
+    - {{ PYTHON }} -m pip install . -vv --config-settings=cmake.args=-DCUDAToolkit_ROOT=/usr/local/cuda
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
```
jakirkham, line 25: Would add the CUDA compiler here

```suggestion
    - {{ compiler('cxx') }}
    - {{ compiler('cuda') }}
```

conda/recipes/pynvjitlink/conda_build_config.yaml
```diff
@@ -0,0 +1,8 @@
+c_compiler_version:
+  - 11
+
+cxx_compiler_version:
+  - 11
+
+cuda_compiler:
+  - cuda-nvcc
```
jakirkham, line 8: This will ensure we use the latest version of CUDA 12 available

```suggestion
cuda_compiler:
  - cuda-nvcc

cuda_compiler_version:
  - 12
```

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+    - cmake >=3.26.4
+    - ninja
+    - sysroot_{{ target_platform }} 2.17
+  host:
+    # We get libnvjitlink from the host's RPM packages, which are updated
+    # sooner than conda sources.
+    #- libnvjitlink-dev
+    #- libnvjitlink-static
```
jakirkham, line 33:
```suggestion
    - libnvjitlink-dev
    - libnvjitlink-static
```

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+    #- cuda-version ={{ cuda_version }}
```
jakirkham, line 34:
```suggestion
    - cuda-version ={{ cuda_compiler_version }}
```

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+    - pynvjitlink/tests/test_nvjitlink_api.py
+    - pynvjitlink/tests/test_nvjitlink.py
+    # Extra files for testing edge cases of libnvjitlink
+    # see test_nvjitlink_api.py for details
+    - pynvjitlink/tests/test_device_functions.cubin
+    - pynvjitlink/tests/test_device_functions.fatbin
+    - pynvjitlink/tests/undefined_extern.cubin
+  requires:
+    - {{ compiler('cuda') }}
+    - cuda-version ={{ cuda_version }}
```
jakirkham, line 56:
```suggestion
    - cuda-version ={{ cuda_compiler_version }}
```

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) 2018-2023, NVIDIA CORPORATION.
+
+{% set py_version = environ['CONDA_PY'] %}
+{% set cuda_version = '.'.join(environ['RAPIDS_CUDA_VERSION'].split('.')[:2]) %}
+{% set cuda_major = cuda_version.split('.')[0] %}
```
jakirkham, line 5: This is always 12 right? If so, think we can drop this and just set it in `conda_build_config.yaml` to use everywhere else

```suggestion
```

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) 2018-2023, NVIDIA CORPORATION.
+
+{% set py_version = environ['CONDA_PY'] %}
```
jakirkham, line 3: This appears to be unused. So lets drop it

```suggestion
```

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+{% set cuda_version = '.'.join(environ['RAPIDS_CUDA_VERSION'].split('.')[:2]) %}
+{% set cuda_major = cuda_version.split('.')[0] %}
+{% set date_string = environ['RAPIDS_DATE_STRING'] %}
```
jakirkham, line 6: Recommending dropping unused variable

```suggestion
```

jakirkham (commented):
In talking offline, our understanding is CUDA 12.2u1 has the memory corruption bug fix we need. The conda-forge packages are based on CUDA 12.2u2 so should also have that fix

Have noted a few suggestions below that would allow us to use CUDA 12.2u2. There may be more changes needed depending. Though hopefully this makes working with the build tooling (like CMake) less complex

We will of course want to test whatever package we build. Am assuming we have some way to do that (either on CI or offline)

pyproject.toml
jakirkham: Think that is a different bug. My understanding is the memory corruption bug was fixed in 12.2u1 whereas conda-forge packages are on 12.2u2

bdice: @jakirkham, you are correct. The bug ids are 4080381 and 4121608. 12.2u1 should have a fix.

ci/build_conda.sh
brandon-b-miller: While discussing offline, @bdice and I realized the following case can occur. Suppose a user has:
- 12.0 Driver
- 12.2 cuDF
- 12.2u2 pynvjitlink
- 12.3 CTK
In this case even with the presence of `pynvjitlink` we will still encounter a failure since there's a piece of the PTX that the 12.3 CTK generates which will be too new for `libnvjitlink` from `12.2u2`. While conda provides some tools to constrain such that this would never happen, the same is not the case for pip packages. This provides some motivation to pursue the RPM package route.

pyproject.toml
bdice: Due to cases [like this](https://github.com/rapidsai/pynvjitlink/pull/33#discussion_r1446404877) we actually do need to build with 12.3 (12.latest in general). The bug fixes may be irrelevant since we have to use 12.3 (or 12.latest) to provide a proper compatibility story with pynvjitlink.

pyproject.toml
```diff
@@ -10,7 +10,7 @@ requires = ["scikit-build-core"]
 build-backend = "scikit_build_core.build"
 
 [project]
-name = "pynvjitlink-cu12"
+name = "pynvjitlink"
```
bdice, line 13: I changed this so that the package name would align for conda. I modified `build_wheel.sh` and `test_wheel.sh` accordingly, so they add a `-cu12` suffix. This matches our "normal" approach for RAPIDS wheels.

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - cuda-version {{ cuda_version }}
+    - python
+    - pip
+    - scikit-build-core
+  run:
+    - python
+    - numba >=0.57
+    - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
+
+#test:
```
bdice, line 37: I struggled with getting conda-build to run tests properly. Not even the import worked. However, I was successful in importing the "broken" package manually and running tests locally in `test_conda.sh` so I'm inclined to remove these recipe tests.

ci/build_wheel.sh
```diff
@@ -6,9 +6,20 @@ set -euo pipefail
...
+
+# Patch project metadata files to include the CUDA version suffix and version override.
+pyproject_file="${package_dir}/pyproject.toml"
+
+sed -i "s/^name = \"${package_name}\"/name = \"${package_name}${PACKAGE_CUDA_SUFFIX}\"/g" ${pyproject_file}
+
 rapids-logger "Build wheel"
-mkdir -p ./wheel-build
-pip wheel . --wheel-dir=./wheel-build -vvv
+mkdir -p ./dist
```
bdice, line 21: I renamed this to `dist` to align with other RAPIDS packages that use this directory name for wheels.

CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ find_package(Python COMPONENTS Interpreter Development REQUIRED)
 Python_add_library(_nvjitlinklib MODULE pynvjitlink/_nvjitlinklib.cpp WITH_SOABI)
 
 find_package(
-  CUDAToolkit 12.3 REQUIRED
+  CUDAToolkit 12.2.1 REQUIRED
```
bdice, line 14: This version should allow conda-forge packages of 12.2u2 to be used. I'm not sure if this is the exact right pinning though, perhaps it reports itself as something like `12.2.140` to CMake and uses the third component from nvcc rather than using the update version. This should be checked before merging.

bdice: Fixing this to 12.2.140 so we require 12.2u2 (12.2u1 might be sufficient but there's no harm in requiring the version on conda-forge since it's newer).

bdice:
I made some good progress and have conda builds working.

I see CI has one test failure on conda for `test_add_fatbin_with_cubin_error`, which I also observe when testing locally. Could this be an issue specific to using nvjitlink from CUDA 12.2u2? Hopefully @gmarkall or @brandon-b-miller know how to diagnose this, since I am not familiar with the internals or tests.

https://github.com/rapidsai/pynvjitlink/actions/runs/7470027483/job/20328142813#step:7:670
```
=================================== FAILURES ===================================
_______________________ test_add_fatbin_with_cubin_error _______________________

device_functions_cubin = b'\x7fELF\x02\x01\x013\x07\x00\x00\x00\x00\x00\x00\x00\x01\x00\xbe\x00z\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x0...00\x00\x00\x00\x00\x00\x03\x00\x00\x00\x07\x00\x00\x18\x80\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'

    def test_add_fatbin_with_cubin_error(device_functions_cubin):
        nvjitlinker = NvJitLinker("-arch=sm_75")
        name = "test_device_functions.cubin"
>       with pytest.raises(NvJitLinkError, match="NVJITLINK_ERROR_INVALID_INPUT error"):
E       Failed: DID NOT RAISE <class 'pynvjitlink.api.NvJitLinkError'>

pynvjitlink/tests/test_pynvjitlink_api.py:100: Failed
```

brandon-b-miller:
> I see CI has one test failure on conda for `test_add_fatbin_with_cubin_error`, which I also observe when testing locally. 

When I bumped from cuda 12.2 to cuda 12.3, this expected failure turned into a pass. It's a loose end from [this conversation](https://github.com/rapidsai/pynvjitlink/pull/10/files#r1394958074). My guess is this was an update to the underlying lib in 12.3.

jakirkham:
Would it make sense to mark it as xfail for now?

bdice:
I xfail'd that test and opened this PR for review.

CMakeLists.txt
```diff
@@ -11,7 +11,8 @@ find_package(Python COMPONENTS Interpreter Development REQUIRED)
 Python_add_library(_nvjitlinklib MODULE pynvjitlink/_nvjitlinklib.cpp WITH_SOABI)
 
 find_package(
-  CUDAToolkit 12.3 REQUIRED
+  # Require CUDA 12.2 Update 2 to avoid nvjitlink bugs
+  CUDAToolkit 12.2.140 REQUIRED
```
gmarkall, line 15: Why does this move backwards from 12.3 to 12.2? I have read the discussions on the PR, but it looked like the conclusions suggested the latest was necessary. (FWIW, my POV has always been that the optimal thing to do is to always build against the latest released version)

gmarkall: If the user has a 12.3 toolkit and 12.2 driver then it wouldn't work. A couple of other references from comments on this PR:

* https://github.com/rapidsai/pynvjitlink/pull/33#discussion_r1446404877
* https://github.com/rapidsai/pynvjitlink/pull/33#discussion_r1446523241

jakirkham: In CUDA 12, the entire CUDA Toolkit is shipped via Conda and there are no CUDA 12.3 Conda packages currently. So that doesn't seem like a risk

We can of course add a `cuda-version` upper bound to guard against this ( https://github.com/rapidsai/pynvjitlink/pull/33/files#r1445457742 )

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,47 @@
+# Copyright (c) 2018-2024, NVIDIA CORPORATION.
+
+{% set data = load_file_data("pyproject.toml") %}
```
jakirkham, line 3:
```suggestion
{% set data = load_file_data("pyproject.toml") %}

{% set cuda_major = cuda_compiler_version.split(".")[0]|int %}
{% set cuda_minor = cuda_compiler_version.split(".")[0]|int %}
```

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - libnvjitlink-dev
+    - libnvjitlink-static
+    - cuda-version {{ cuda_compiler_version }}
+    - python
+    - pip
+    - scikit-build-core
+  run:
+    - python
+    - numba >=0.57
+    - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
```
jakirkham, line 37:
```suggestion
    - {{ pin_compatible('cuda-version', min_pin='x', max_pin=(cuda_major ~ "." ~ (cuda_minor + 1))) }}
```

CMakeLists.txt
jakirkham: The other thing to keep in mind is there will always be newer CUDA Toolkits released after we have published RAPIDS releases (especially when considering LTS releases). So we cannot guarantee users will have a version of `pynvjitlink` that was built against the just recently shipped CUDA Toolkit

Think the best we can do in those cases is have proper version constraints and maybe do some validation on `import pynvjitlink` to catch any other aberrant cases

gmarkall: > In CUDA 12, the entire CUDA Toolkit is shipped via Conda and there are no CUDA 12.3 Conda packages currently. So that doesn't seem like a risk

The user could install Numba with conda, not install the CUDA toolkit at all via conda, and use a system-installed 12.3 toolkit - you don't need to conda-install a CUDA toolkit to use it with Numba.

jakirkham: At least today, that is not a supported use case of CUDA with Conda in RAPIDS or with conda-forge packages

bdice: @gmarkall We discussed this topic a bit on Slack. I'll tag you there. This PR implements a "short term plan" by using conda-forge's pynvjitlink from 12.2u2 so we can update RAPIDS to build with CUDA 12.2. We've also discussed a "long term plan" to use tarballs or some other upstream source that reliably provides "latest CUDA" that we can use as a single source for both conda and pip packages. That will align more closely with what you're seeking. For this PR, we just need to identify if this is good enough for a first iteration of pynvjitlink conda packages. This is needed to unblock cudf building with CUDA 12.2, since pynvjitlink will be a hard dependency of cudf.

.github/actions/compute-matrix/action.yaml
```diff
@@ -33,12 +35,26 @@ runs:
           - { CUDA_VER: '12.0.1', ARCH: 'arm64', PY_VER: '3.11', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
         "
 
+        export TEST_MATRIX_CONDA="
+          - { CUDA_VER: '12.0.1', ARCH: 'amd64', PY_VER: '3.9', LINUX_VER: 'ubuntu20.04', gpu: 'v100', driver: 'latest' }
+          - { CUDA_VER: '12.0.1', ARCH: 'amd64', PY_VER: '3.10', LINUX_VER: 'ubuntu20.04', gpu: 'v100', driver: 'latest' }
+          - { CUDA_VER: '12.0.1', ARCH: 'amd64', PY_VER: '3.11', LINUX_VER: 'ubuntu20.04', gpu: 'v100', driver: 'latest' }
+          - { CUDA_VER: '12.0.1', ARCH: 'arm64', PY_VER: '3.9', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
+          - { CUDA_VER: '12.0.1', ARCH: 'arm64', PY_VER: '3.10', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
+          - { CUDA_VER: '12.0.1', ARCH: 'arm64', PY_VER: '3.11', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
```
vyasr, line 44: What's the rationale for having two separate test matrices? AFAICT the only difference is that on x86 wheels test on Ubuntu 18.04 while conda tests on 20.04. Why can't/shouldn't we unify them?

ci/build_wheel.sh
```diff
@@ -6,9 +6,18 @@ set -euo pipefail
...
+PACKAGE_CUDA_SUFFIX="-${RAPIDS_PY_CUDA_SUFFIX}"
+
+# Patch project metadata files to include the CUDA version suffix.
+sed -i "s/^name = \"pynvjitlink\"/name = \"pynvjitlink${PACKAGE_CUDA_SUFFIX}\"/g" pyproject.toml
+
 rapids-logger "Build wheel"
-mkdir -p ./wheel-build
-pip wheel . --wheel-dir=./wheel-build -vvv
+mkdir -p ./dist
+python -m pip wheel . --wheel-dir=./dist -vvv --disable-pip-version-check
```
vyasr, line 20: This change is mostly for future-proofing since pynvjitlink doesn't have any dependencies right now.
```suggestion
python -m pip wheel . --wheel-dir=./dist -vvv --disable-pip-version-check --no-deps
```

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - {{ PYTHON }} -m pip install . -vv
+  ignore_run_exports_from:
+    - libnvjitlink-dev
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ compiler('cuda') }}
+    - cuda-version {{ cuda_compiler_version }}
```
vyasr, line 23: Right now this is basically set so that we'll always use the latest major version available on cf. Since this is our short-term solution until we figure out how to use non-conda packages to get the latest CTK, should we be pinning more precisely for now to exactly the build we want, or do we want to let this float across minor versions?

conda/recipes/pynvjitlink/meta.yaml
vyasr: Is this valid? I thought `pin_compatible` only supported pinning expressions (i.e. literal strings of the form `x[.x[.x...]]`, not actual versions. Maybe we can use the `upper_bound` parameter for what you're trying to do here? https://docs.conda.io/projects/conda-build/en/stable/resources/variants.html#extra-jinja2-functions

pynvjitlink/tests/test_pynvjitlink_api.py
```diff
@@ -94,6 +94,11 @@ def test_add_cubin_with_fatbin_error(device_functions_fatbin):
         nvjitlinker.add_cubin(device_functions_fatbin, name)
 
 
+@pytest.mark.xfail(
+    reason="Fails in conda due to using CUDA 12.2u2. This should be fixed "
+    "when requiring CUDA 12.3 nvjitlink library. See: "
+    "https://github.com/rapidsai/pynvjitlink/pull/33#issuecomment-1885182905"
+)
```
vyasr, line 101: The inability to write this conditionally makes me wonder if pynvjitlink should embed the CUDA version used at build-time as a compile-time constant accessible via a Python API. That would be very helpful for us for debugging in the future. Since we're statically linking our only recourse for debugging purposes right now is to check when a particular pynvjitlink package was built and then see what version of CUDA the build scripts used at that time, right?

I'm not proposing a fix in this PR since that's out of scope, but IIUC that would be a nice follow-up.

vyasr (changes requested):

.github/actions/compute-matrix/action.yaml
bdice: I am okay with unifying them. I created a separate test matrix for conda because we can't test 18.04 with conda (no images for that) and didn't want to change the wheel test matrix.

ci/build_wheel.sh
bdice: I don't know what this flag is meant to do. Is this command doing an isolated build? Does it skip installing the "build system requirements" when this is specified?

conda/recipes/pynvjitlink/meta.yaml
bdice: I'm happy to let it float across minor versions until we can separate this from conda-forge entirely, per the ["long term plan."](https://github.com/rapidsai/pynvjitlink/pull/33#discussion_r1449272016)

pynvjitlink/tests/test_pynvjitlink_api.py
bdice: This would be fine with me but I don't think it's in scope here. It'll also be ameliorated somewhat by the ["long term plan"](https://github.com/rapidsai/pynvjitlink/pull/33#discussion_r1449272016) to build both pip and conda packages with a single nvjitlink version, which will establish a correspondence between pynvjitlink versions and nvjitlink versions.

ci/build_wheel.sh
vyasr: It skips run requirements. `pip wheel` downloads runtime dependencies by default. The build is isolated by default, yes. There is of course no way to turn off downloading build dependencies when doing an isolated build since that would always be broken.

pynvjitlink/tests/test_pynvjitlink_api.py
vyasr: I agree that it's not in scope here. The long term plan doesn't really help much though, except that in the long term we will hopefully always support the latest CUDA version so we can always ask users to update to the latest pynvjitlink version and hope that fixes their problems. But for debugging purposes in the context of a broader error it would help us narrow down issues in bug reports.

.github/actions/compute-matrix/action.yaml
vyasr: Gotcha. For the purpose of verifying that we're making manylinux wheels properly I'm fine testing on older systems at the cost of keeping two separate matrices.

conda/recipes/pynvjitlink/meta.yaml
bdice: Yes, I think we want `upper_bound`. Fixed in 108e2bf0b7bd241316cce69840a732521554f4d9.

.github/actions/compute-matrix/action.yaml
bdice: Unified the test matrices and dropped Ubuntu 18.04 in 5a27d41bb71f5cac728bb0a8d83e9d91ef267dc6.

conda/recipes/pynvjitlink/meta.yaml
jakirkham: Actually as `{{ compiler('cuda') }}` already comes with `cuda-version`, we could drop this from `build`

```suggestion
```

pynvjitlink/tests/test_pynvjitlink_api.py
jakirkham: Actually am hoping we not only provide the ability to do this kind of check, but we actually do the check as part of `import pynvjitlink` (and error if it is not compatible)

It's possible that won't be enough and we will need to add some runtime checks for when `pynvjitlink` is used (and even defer some cases to the user to check)

vyasr: Opened https://github.com/rapidsai/pynvjitlink/issues/38 to continue this discussion separately.

.github/actions/compute-matrix/action.yaml
vyasr: Also OK with me.

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+{% if cuda_compiler_version is defined %}
+{% set cuda_major = cuda_compiler_version.split(".")[0]|int %}
+{% set cuda_minor = cuda_compiler_version.split(".")[1]|int %}
+{% else %}
+{% set cuda_major = 0 %}
+{% set cuda_minor = 0 %}
+{% endif %}
+
+package:
+  name: pynvjitlink
```
vyasr, line 14:
```suggestion
  name: {{ project_data["name"] }}
```

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+about:
+  home: https://rapids.ai/
+  dev_url: https://github.com/rapidsai/pynvjitlink
+  doc_url: https://github.com/rapidsai/pynvjitlink/blob/main/README.md
+  license: Apache-2.0
+  license_family: Apache
+  license_file: LICENSE
+  license_url: https://github.com/rapidsai/pynvjitlink/blob/main/LICENSE
+  summary: nvJitLink Python binding
```
vyasr, line 55:
```suggestion
  summary: {{ project_data["description"] }}
```

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+{% set cuda_major = cuda_compiler_version.split(".")[0]|int %}
+{% set cuda_minor = cuda_compiler_version.split(".")[1]|int %}
+{% else %}
+{% set cuda_major = 0 %}
+{% set cuda_minor = 0 %}
+{% endif %}
+
+package:
+  name: pynvjitlink
+  version: {{ data.get("project", {}).get("version", "") }}
```
vyasr, line 15:
```suggestion
  version: {{ project_data["version"] }}
```

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+  run:
+    - python
+    - numba >=0.57
+    - {{ pin_compatible('cuda-version', min_pin='x', upper_bound=(cuda_major ~ "." ~ (cuda_minor + 1))) }}
+
+about:
+  home: https://rapids.ai/
+  dev_url: https://github.com/rapidsai/pynvjitlink
+  doc_url: https://github.com/rapidsai/pynvjitlink/blob/main/README.md
+  license: Apache-2.0
```
vyasr, line 51: I don't remember whether we want "Apache 2.0" or "Apache-2.0" (I know there is some context where there is a difference), but I imagine that we want the same in both pip and conda packages so we should make sure that what pyproject.toml has is correct. In case the two actually need to be different for some reason, then ignore this change.
```suggestion
  license: {{ project_data["license"] }}
```

conda/recipes/pynvjitlink/conda_build_config.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+cuda_compiler_version:
+  - 12
```
jakirkham, line 11: Think we may need to fill in the minor version part so we can access that in the recipe. Perhaps there's a better solution than doing this

```suggestion
  - 12.2
```

pyproject.toml
vyasr, line 1: Please add the following section:
```
[project.urls]
Homepage = "https://rapids.ai/"
Documentation = "https://github.com/rapidsai/pynvjitlink/blob/main/README.md"
Repository = "https://github.com/rapidsai/pynvjitlink"
License = "https://github.com/rapidsai/pynvjitlink/blob/main/LICENSE"
```

Then use that in the conda recipe.

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+    - python
+    - pip
+    - scikit-build-core
+  run:
+    - python
+    - numba >=0.57
+    - {{ pin_compatible('cuda-version', min_pin='x', upper_bound=(cuda_major ~ "." ~ (cuda_minor + 1))) }}
+
+about:
+  home: https://rapids.ai/
```
vyasr, line 48:
```suggestion
  home: {{ project_urls["Homepage"] }}
```

conda/recipes/pynvjitlink/conda_build_config.yaml
jakirkham: Without this am getting errors like...

```
$ conda render conda/recipes/pynvjitlink
Error: Failed to render jinja template in /Users/jkirkham/Developer/rapidsai/pynvjitlink/conda/recipes/pynvjitlink/meta.yaml:
list object has no element 1
```

..., which I'm reading as this line failing when `cuda_compiler_version = "12"`

https://github.com/rapidsai/pynvjitlink/blob/cc57af8de24e7101f163d6f96931e05e9ab49ac3/conda/recipes/pynvjitlink/meta.yaml#L7

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,55 @@
+# Copyright (c) 2018-2024, NVIDIA CORPORATION.
+
+{% set data = load_file_data("pyproject.toml") %}
```
vyasr, line 3: Let's assume all these fields exist. I would rather fail than embed empty versions (or any other metadata).
```suggestion
{% set data = load_file_data("pyproject.toml") %}
{% set project_data = data["project"] %}
{% set project_urls = project_data["urls"] %}
```

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+    - pip
+    - scikit-build-core
+  run:
+    - python
+    - numba >=0.57
+    - {{ pin_compatible('cuda-version', min_pin='x', upper_bound=(cuda_major ~ "." ~ (cuda_minor + 1))) }}
+
+about:
+  home: https://rapids.ai/
+  dev_url: https://github.com/rapidsai/pynvjitlink
```
vyasr, line 49:
```suggestion
  dev_url: {{ project_urls["Repository"] }}
```

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+  doc_url: https://github.com/rapidsai/pynvjitlink/blob/main/README.md
```
vyasr, line 50:
```suggestion
  doc_url: {{ project_urls["Documentation"] }}
```

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+  license: Apache-2.0
+  license_family: Apache
+  license_file: LICENSE
+  license_url: https://github.com/rapidsai/pynvjitlink/blob/main/LICENSE
```
vyasr, line 54:
```suggestion
  license_url: {{ project_urls["License"] }}
```

vyasr (changes requested):
Since we're already reading from pyproject.toml we may as well read as much metadata as we can from there.

conda/recipes/pynvjitlink/conda_build_config.yaml
bdice: Just saw this, too -- fixed in b5d04c3.

conda/recipes/pynvjitlink/meta.yaml
bdice: conda-build fails with this change:
```
Error: Failed to render jinja template in /pynvjitlink/conda/recipes/pynvjitlink/meta.yaml:
Encountered unknown tag 'project_data'.
```

We really do have to use the `.get(...).get(...)` as far as I can tell.

conda/recipes/pynvjitlink/meta.yaml
bdice: How would we be setting the cuda compiler version without this line?

conda/recipes/pynvjitlink/meta.yaml
jakirkham: Yeah that was what we did [in Dask-CUDA]( https://github.com/rapidsai/dask-cuda/blob/d565a17f0c255eca5b9d3d8493fb26d2c77aedc1/conda/recipes/dask-cuda/meta.yaml#L49-L58 )

conda/recipes/pynvjitlink/meta.yaml
jakirkham: Via `cuda_compiler_version` in `conda_build_config.yaml`. This is how it [works in conda-forge]( https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/cbccb41e313c017c1ad15db6c287e87aa57d54dd/recipe/conda_build_config.yaml#L44-L47 ) as well

bdice: Gotcha. Then yes, I'll remove this line.

conda/recipes/pynvjitlink/meta.yaml
vyasr: It seems like the name cannot be templated, likely due to some issue with evaluation order of Jinja variables in conda build.

conda/recipes/pynvjitlink/meta.yaml
vyasr: This worked fine for me locally. Perhaps some typo in how you copied it down? I've pushed my changes.

vyasr (approved):

jakirkham (approved):

raydouglass (approved):

jakirkham:
Thanks all! 🙏

Let's follow up on anything else separately

.github/workflows/build.yaml
```diff
@@ -28,6 +28,14 @@ jobs:
       build_type: branch
       script: "ci/build_wheel.sh"
       matrix_filter: ${{ needs.compute-matrix.outputs.BUILD_MATRIX }}
+  build-conda:
+    needs:
+      - compute-matrix
+    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@branch-24.02
+    with:
+      build_type: pull-request
```
ajschmidt8, line 36: this should be `build_type: ${{ inputs.build_type || 'branch' }}`.

also, it will need the following:

```yaml
branch: ${{ inputs.branch }}
date: ${{ inputs.date }}
sha: ${{ inputs.sha }}
```

jakirkham: This is the `build-conda` portion, which IIUC is just for testing the packages do build

https://github.com/rapidsai/pynvjitlink/blob/a1479d31c23120b65532e8f31f20b64a69e19d5c/.github/workflows/build.yaml#L31-L38

AFAICT this matches what the `build-wheels` section has

https://github.com/rapidsai/pynvjitlink/blob/a1479d31c23120b65532e8f31f20b64a69e19d5c/.github/workflows/build.yaml#L23-L30

<hr>

The `publish-conda` section has these changes already

https://github.com/rapidsai/pynvjitlink/blob/a1479d31c23120b65532e8f31f20b64a69e19d5c/.github/workflows/build.yaml#L50-L59

It appears similar to `publish-wheels`, which also has them

https://github.com/rapidsai/pynvjitlink/blob/a1479d31c23120b65532e8f31f20b64a69e19d5c/.github/workflows/build.yaml#L39-L49

<hr>

Based on this, do we have what we need already?

Or do both Conda & Wheels sections need some updates?

jakirkham: That said, it is odd that `build-conda` has `build_type: pull-request` instead of `build_type: branch` like in `build-wheels`. Fixing with PR ( https://github.com/rapidsai/pynvjitlink/pull/43 )

jakirkham: It looks like the `build_type` change noted above was all that was needed. Though please let us know if we are missing something

jakirkham:
Now that conda-forge has CUDA 12.3, have submitted PR ( https://github.com/rapidsai/pynvjitlink/pull/50 ) to update to CUDA 12.3. Hopefully that closes out a few threads above

PR #44, opened 2024-01-16
Release 0.1.9

jakirkham:
Fixes for previous botched releases

* Had tagged `0.1.8`, but missed these changes. Fixing by bumping past it
* Also missed that `pynvjitlink/VERSION` has a hard-coded version in `0.1.7`. So updating it now

Would recommend we simplify/automate this release process


brandon-b-miller (approved):

jakirkham:
Thanks Bradley and Brandon! 🙏

PR #54, opened 2024-02-13
Use add_library to link archives in Numba patch

### LINKED ISSUE #53: [BUG] Patching breaks cooperative groups
Patching numba seems to disable launching kernels that involve cooperative groups:

```python
from numba import cuda
from pynvjitlink import patch

patch.patch_numba_linker()

@cuda.jit
def cg_kernel():
    grid = cuda.cg.this_grid()

cg_kernel[1,1]()
```
this produces the following error from [here](https://github.com/rapidsai/pynvjitlink/blob/main/pynvjitlink/patch.py#L107)
```
numba.cuda.cudadrv.driver.LinkerError: Don't know how to link archives
```
Using numba `0.59.0`.

gmarkall:
The required functionality has been available in the underlying API, and just not called by the patch. This enables linking of archives by the patched linker, and therefore fixes the issues with cooperative groups in Numba.

Fixes #53

### DIFF
--- a/pynvjitlink/patch.py
+++ b/pynvjitlink/patch.py
@@ -104,7 +104,7 @@ class PatchedLinker(Linker):
         elif kind == FILE_EXTENSION_MAP["fatbin"]:
             fn = self._linker.add_fatbin
         elif kind == FILE_EXTENSION_MAP["a"]:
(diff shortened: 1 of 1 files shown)


jakirkham:
Thanks Graham and Brandon! 🙏

PR #56, opened 2024-02-14
Show config info in tests

jakirkham:
* Run `nvidia-smi` in wheel tests to confirm GPUs are working
* Display Numba's system information

### DIFF
--- a/ci/test_wheel.sh
+++ b/ci/test_wheel.sh
@@ -19,6 +19,9 @@ pushd test_binary_generation
 make
 popd
 
+rapids-logger "Check GPU usage"
+nvidia-smi
+
 rapids-logger "Run Tests"
 pip install pytest
 pytest pynvjitlink/tests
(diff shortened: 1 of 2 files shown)


jakirkham:
From the Conda tests, it looks like Numba can't find `nvrtc`:

```
Finding nvrtc from Conda environment (NVIDIA package)
	Located at libnvrtc.so
	Trying to open library...	ERROR: failed to open nvrtc:
libnvrtc.so: cannot open shared object file: No such file or directory
```

Though it does find `libcudadevrt.a`, which had been a concern

```
Finding cudadevrt from Conda environment (NVIDIA package)
	Located at /opt/conda/envs/test/lib/libcudadevrt.a
	Checking library...	ok
```

ci/test_wheel.sh
```diff
@@ -19,6 +19,12 @@ pushd test_binary_generation
 make
 popd
 
+rapids-logger "Check GPU usage"
+nvidia-smi
+
+rapids-logger "Show Numba CUDA library configuration"
+python -c "from numba.cuda.cudadrv.libs import test; test()"
+
```
jakirkham, line 27: Do we need to do something differently with the wheel tests? It can't find Numba

```python
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'numba'
```

jakirkham: Looks like Numba is just not getting installed. Added to PR: https://github.com/rapidsai/pynvjitlink/pull/57

jakirkham:
Folding into PR: https://github.com/rapidsai/pynvjitlink/pull/57

PR #55, opened 2024-02-14
Test with cuda-cudart-static libraries for cooperative group support

bdice:
In https://github.com/rapidsai/pynvjitlink/issues/53#issuecomment-1942948977, @gmarkall noticed that some Numba tests for cooperative groups were being skipped. They show a skip reason of `cudadevrt missing`. I traced [this Numba test skipping code](https://github.com/numba/numba/blob/3edb458c69daeaaffbfc938145c367eda1cf043f/numba/cuda/testing.py#L170-L182) and the dependency list in `test_patch.sh`, and found that `libcudadevrt.a` was probably not available in the CI environment.

This PR switches from `cuda-nvcc-impl` to `cuda-nvcc` in the CI environment, which includes the missing `libcudadevrt.a`. In particular, this will ensure we get `cuda-cudart-static` packages in the testing environment.

We can verify this PR by ensuring that the following tests are run (and not skipped):
```
test_cache_cg (numba.cuda.tests.cudapy.test_caching.CUDACachingTest.test_cache_cg) ... skipped 'cudadevrt missing'
...
test_false_cooperative_doesnt_link_cudadevrt (numba.cuda.tests.cudapy.test_cooperative_groups.TestCudaCooperativeGroups.test_false_cooperative_doesnt_link_cudadevrt)
We should only mark a kernel as cooperative and link cudadevrt if the ... skipped 'cudadevrt missing'
test_max_cooperative_grid_blocks (numba.cuda.tests.cudapy.test_cooperative_groups.TestCudaCooperativeGroups.test_max_cooperative_grid_blocks) ... skipped 'cudadevrt missing'
test_sync_at_matrix_row (numba.cuda.tests.cudapy.test_cooperative_groups.TestCudaCooperativeGroups.test_sync_at_matrix_row) ... skipped 'cudadevrt missing'
test_sync_group (numba.cuda.tests.cudapy.test_cooperative_groups.TestCudaCooperativeGroups.test_sync_group) ... skipped 'cudadevrt missing'
test_sync_group_is_cooperative (numba.cuda.tests.cudapy.test_cooperative_groups.TestCudaCooperativeGroups.test_sync_group_is_cooperative) ... skipped 'cudadevrt missing'
test_this_grid (numba.cuda.tests.cudapy.test_cooperative_groups.TestCudaCooperativeGroups.test_this_grid) ... skipped 'cudadevrt missing'
[...]

PR #57, opened 2024-02-14
Install missing test deps & align testing protocols across formats

jakirkham:
Install a few things that were needed

* `cuda-nvrtc` for Conda tests
* Use `cuda-nvcc` in all cases
* Wheel dependencies at start (`numba` was missed; `pytest` was done later)
* Also run with `python -m <...>` to ensure the same `python` is used for installs & tests
* Run `nvidia-smi` in all tests (missing in some places)
* Run `python -m numba --sysinfo` in all tests (gets info about Numba's config and what Numba is missing)
* Install `psutil` for more Numba config info
* Generally align testing across package formats

Fixes https://github.com/rapidsai/pynvjitlink/pull/55

### DIFF
--- a/ci/test_conda.sh
+++ b/ci/test_conda.sh
@@ -41,6 +41,9 @@ popd
 rapids-logger "Check GPU usage"
 nvidia-smi
 
+rapids-logger "Show Numba system info"
+python -m numba --sysinfo
+
 EXITCODE=0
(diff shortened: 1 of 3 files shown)


bdice (approved):

jakirkham:
Thanks Bradley! 🙏

PR #58, opened 2024-02-16
Release 0.1.13 & use `VERSION` file only

jakirkham:
* Update to 0.1.13
* Use `VERSION` file to set version dynamically

xref: https://github.com/rapidsai/build-planning/issues/15


CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.26.4 FATAL_ERROR)
 
 project(
   pynvjitlink
-  VERSION 0.1.12
+  VERSION ${SKBUILD_PROJECT_VERSION}
```
jakirkham, line 5: `scikit-build-core` [sets `${SKBUILD_PROJECT_VERSION}`]( https://scikit-build-core.readthedocs.io/en/latest/cmakelists.html#accessing-information ) for us

So just use that here

pyproject.toml
```diff
@@ -28,6 +28,3 @@ License = "https://github.com/rapidsai/pynvjitlink/blob/main/LICENSE"
 
 [tool.setuptools]
 license-files = ["LICENSE"]
-
-[tool.setuptools.dynamic]
-version = {file = "pynvjitlink/VERSION"}
```
jakirkham, line 33: AFAICT this was unused. Though please let me know if I'm missing something

pynvjitlink/VERSION
```diff
@@ -1 +0,0 @@
-0.1.12
```
jakirkham, line 1: Dropped this file as it appeared to be unused

Perhaps there is some way to have everything [dynamically use this version with `scikit-build-core`]( https://scikit-build-core.readthedocs.io/en/latest/configuration.html#dynamic-metadata ), but wasn't quite seeing how to get that to work without adding more stuff. So just stuck with `pyproject.toml`

ci/release/update-version.sh
```diff
@@ -16,6 +16,4 @@ function sed_runner() {
 }
 
 # Centralized version file update
-echo "${NEXT_FULL_TAG}" > pynvjitlink/VERSION
-sed_runner 's/'"^  VERSION [0-9\.]*"'/'"  VERSION ${NEXT_FULL_TAG}"'/g' CMakeLists.txt
```
jakirkham, line 20: These no longer need updates. So dropped them

ci/release/update-version.sh
```diff
@@ -16,6 +16,4 @@ function sed_runner() {
...
 sed_runner 's/^version = "[0-9\.]*"/version = "'${NEXT_FULL_TAG}'"/g' pyproject.toml
```
jakirkham, line 19: As the version now lives in one place, maybe this script isn't needed?

Left it for now as it could be ergonomic when working across projects. Though open to other thoughts here

bdice: Please leave this script. It helps a lot to have a single reference for how versions are used in every RAPIDS-related project.

pynvjitlink/VERSION
bdice: We should try to adopt the same process as cudf and other libraries that dynamically fetch their version info in pyproject.toml from a VERSION file. It would be ideal for VERSION to be the central source of truth rather than pyproject.toml.

This shows how it works: https://github.com/rapidsai/cudf/blob/aa9d4846e80fad133e7af19aba99cefa04cb8b7b/python/cudf/pyproject.toml#L175

Happy to answer questions on this process if needed.

bdice (changes requested):
Suggested a different approach to version consolidation.

pynvjitlink/VERSION
jakirkham: Thanks Bradley! 🙏

The example helped. Have switched things around to keep the `VERSION` file

ci/release/update-version.sh
jakirkham: Yep it is still there 🙂

Made the same observation above

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -1,12 +1,16 @@
 # Copyright (c) 2018-2024, NVIDIA CORPORATION.
 
 {% set data = load_file_data("pyproject.toml") %}
+{% set version = load_file_regex(
+    load_file="pynvjitlink/VERSION",
+    regex_pattern="(?P<value>.*)"
+)[0] %}
```
jakirkham, line 7: Similar to `scikit-build-core`'s regex, [`conda-build` has a similar feature]( https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#loading-data-from-other-files )

Rewrote the logic above to load from the file `VERSION`

Note this was also need as conda-build can't run `scikit-build-core` first before generating the YAML (and it is probably better that it doesn't). So this reads it directly from the source of truth

conda/recipes/pynvjitlink/meta.yaml
```diff
@@ -1,12 +1,16 @@
...
 {% set project_data = data.get("project") %}
 {% set project_urls = project_data["urls"] %}
 
 package:
   name: pynvjitlink
-  version: {{ project_data["version"] }}
+  version: {{ version }}
```
jakirkham, line 13: The version field is left blank now as it is defined dynamically. Fortunately we loaded it from the file already, so we can just use it here

jakirkham:
Updated based on the feedback above. Also included similar changes to the Conda recipes

Added a few comments above to highlight those changes

Please let me know what you think 🙂

brandon-b-miller (approved):

bdice (approved):

jakirkham:
Thanks all! 🙏

jakirkham:
Have added a new tag: [`v0.1.13`]( https://github.com/rapidsai/pynvjitlink/releases/tag/v0.1.13 )

Release builds are running on [this GHA]( https://github.com/rapidsai/pynvjitlink/actions/runs/7981304677 )

jakirkham:
Packages are now up! 🥳

* [Conda packages]( https://anaconda.org/rapidsai-nightly/pynvjitlink/files?version=0.1.13 )
* [Wheel packages]( https://anaconda.org/rapidsai-wheels-nightly/pynvjitlink-cu12/files?version=0.1.13 )

PR #60, opened 2024-03-01
Numba patch: Support linking code from memory

gmarkall:
Adds support for linking code from memory to Numba's `@cuda.jit` decorator (in addition to the already-supported linking files from disk). New classes are added to Numba's top-level `cuda` module:

* `Archive`: An archive of objects
* `CUSource`: A CUDA C/C++ source
* `Cubin`: A cubin ELF
* `Fatbin`: A fatbin ELF
* `Object`: An object file
* `PTXSource`: PTX assembly source code.

These are all used by constructing them with a single argument, the code in memory to use. Once created, they can then be passed to the `link=` kwarg of the `@cuda.jit` decorator. An example showing a use case with a CUDA C/C++ source is added.

This implementation is inspired by the approach outlined by @ed-o-saurus in numba/numba#9470.

Notes on changes:
* Various tests now run on the GPU, so test binaries need to be generated using the relevant compute capability - this change is applied in the `Makefile`, along with some refactoring to tidy it up a little.
* Test for new functionality are added. In addition, existing tests that used the test binaries are all modified such that they use a relevant compute capability (usually the one in the test machine, for most of them) because the test binaries are now built for the CC of the current GPU - it's no longer sufficient to hard code CCs like 7.5 or 7.0 for tests. Since fixtures are needed across test files, I've started moving them into `conftest.py`.


ed-o-saurus:
I like this. It's a lot broader than my PR which only supported PTX code. 
One thought: Is it a good idea to name a class `Object`? It could easily be confused with python's `object` class.

gmarkall:
> One thought: Is it a good idea to name a class `Object`? It could easily be confused with python's `object` class.

It would generally be referred to in any user code as `cuda.Object` - does that allay your concern?

gmarkall:
> I like this. It's a lot broader than my PR which only supported PTX code.

Many thanks for your thoughts here. Although this is going into pynvjitlink now, my intention would be to then upstream this into Numba once it's shown to work OK in use in pynvjitlink. Would you be OK with me doing that with this code, with appropriate credit to you for the initial idea / implementation in the Numba changelog and release notes?

I hope you don't mind that I took this and ran with it this far - it's somewhat on the critical path for us, and I feared that describing exactly what I needed to you and then expecting you to jump to implement it would be an unreasonable burden to place on you.

ed-o-saurus:
If users will be instantiating the class as `cuda.Object`, that's a lot clearer. 

By all mean, go ahead and push this to the numba repo. As long as I get a mention in the release notes, I'm happy.

bdice:
@gmarkall The changes to wheel testing were merged in #62, so merging the upstream will reduce them out of the diff here. I'll go ahead and do that, and update the description.

pynvjitlink/patch.py
```diff
@@ -91,6 +147,34 @@ def add_ltoir(self, ltoir, name="<external-ltoir>"):
...
+        # here with the ability to provide a file from memory.
+
+        # To maintain compatibility with the original interface, all strings
+        # are treated as paths in the filesystem.
+        if isinstance(path_or_code, str):
+            # Upstream numba does not yet recognize LTOIR, so handle that
+            # separately here.
+            extension = pathlib.Path(path_or_code).suffix
+            if extension == ".ltoir":
+                self.add_file(path_or_code, 'ltoir')
```
jakirkham, line 162:
```suggestion
                self.add_file(path_or_code, "ltoir")
```

pynvjitlink/patch.py
```diff
@@ -109,6 +196,8 @@ def add_file(self, path, kind):
             return self.add_ptx(data, name)
         elif kind == FILE_EXTENSION_MAP["o"]:
             fn = self._linker.add_object
+        elif kind == 'ltoir':
```
jakirkham, line 199:
```suggestion
        elif kind == "ltoir":
```

jakirkham (commented):
Looks like the style check failed. It noted some recommended changes. Have included them below

pynvjitlink/tests/test_pynvjitlink.py
```diff
@@ -98,16 +98,16 @@ def test_complete_empty():
         ("device_functions_archive", InputType.LIBRARY),
     ],
 )
-def test_add_file(input_file, input_type, request):
```
isVoid, line 101: I'll comment on the nearest line to L93, for ltoir test case, wouldn't you just need to add `-lto` flag to `.create` method?

isVoid (approved):
LGTM, with a few small non-blocking comments.

examples/jit_link_data.py
```diff
@@ -0,0 +1,44 @@
...
+
+from numba import cuda
+from pynvjitlink import patch
+
+import numpy as np
+
+patch.patch_numba_linker()
+
+source = cuda.CUSource(
+    """
```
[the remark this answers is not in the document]
gmarkall: The style checker doesn't like that, unfortunately.

pynvjitlink/tests/test_pynvjitlink.py
gmarkall: Thanks for pointing this out... The input type was also wrong - it's not "raw" LTO-IR, it's wrapped in an ELF, so the input type needed to be `OBJECT`, which nvJitLink then extracts the LTO-IR from. Because it's a little different to the other cases, I've moved it into a separate test to fix up.

jakirkham:
Thanks all! 🙏

Do we want to put this in a new release?

isVoid:
If possible, yes. I would like to use this feature for Numbast.

jakirkham:
Ok could you please run?

```sh
cd <pynvjitlink repo top-level directory>
bash update-version.sh <new_version>
```

Where new version is the next untaken version. Please choose whether this is a bug fix (`0.1.15`) or new user facing feature (`0.2.0`)

Then commit the changes and push to a new PR

isVoid:
Sorry about the delay @jakirkham here's the version bump: https://github.com/rapidsai/pynvjitlink/pull/68

jakirkham:
No worries. Thanks Michael! 🙏

PR #61, opened 2024-03-01
Fix nvjitlink doc link in README.md

isVoid:
The link to nvjitlink in README.md isn't working, this PR fixes that.


gmarkall (approved):
Indeed it wasn't working, thanks for the fix!

PR #62, opened 2024-03-05
Update to CUDA 12.4

bdice:
This PR updates pynvjitlink to use CUDA 12.4. This depends on https://github.com/conda-forge/libnvjitlink-feedstock/pull/9.

### DIFF
--- a/ci/test_wheel.sh
+++ b/ci/test_wheel.sh
@@ -14,15 +14,11 @@ rapids-logger "Download Wheel"
 RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
 RAPIDS_PY_WHEEL_NAME="pynvjitlink_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 ./dist/
 
-# This is the version of the suffix with a preceding hyphen. It's used
-# everywhere except in the final wheel name.
-PACKAGE_CUDA_SUFFIX="-${RAPIDS_PY_CUDA_SUFFIX}"
-
 RAPIDS_TESTS_DIR=${RAPIDS_TESTS_DIR:-"${PWD}/test-results"}/
 mkdir -p "${RAPIDS_TESTS_DIR}"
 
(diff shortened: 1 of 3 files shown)


jakirkham:
Thanks Bradley! 🙏

Think we also need PR: https://github.com/conda-forge/cuda-nvcc-feedstock/pull/37

jakirkham:
Think we have what we need. Trying to restart CI to see if that fixes the issue

jakirkham:
Looks like the Conda builds completed successfully! 🎉

CI is still running

bdice:
Awesome. After this CI run completes, I want to push one more change: ab2ec91af7588ffcab161053e00e5279806d6fc1 from #60, since it fixes a bug in the wheel testing script.

jakirkham:
Is it worth putting that in its own PR?

Happy to merge that fix first

bdice:
I saw enough of the CI output. I pushed the fix for wheel tests so we are testing the actual CI artifacts rather than nightly packages.

ci/test_wheel.sh
```diff
@@ -22,7 +22,7 @@ RAPIDS_TESTS_DIR=${RAPIDS_TESTS_DIR:-"${PWD}/test-results"}/
 mkdir -p "${RAPIDS_TESTS_DIR}"
 
 rapids-logger "Install wheel"
-python -m pip install --find-links ./dist pynvjitlink${PACKAGE_CUDA_SUFFIX}
+python -m pip install $(echo ./dist/pynvjitlink*.whl)
```
jakirkham, line 25: Do we still want the CUDA suffix? Or does that matter?

If we don't need the suffix, should we drop `PACKAGE_CUDA_SUFFIX` from above?

```suggestion
python -m pip install $(echo ./dist/pynvjitlink-${PACKAGE_CUDA_SUFFIX}*.whl)
```

bdice: I pushed a change for this.

jakirkham (approved):
LGTM. Thanks Bradley! 🙏

jakirkham:
Looks like we do this the old school way

Issue #63, opened 2024-03-06
Is `.ycm_extra_conf.py` needed?


jakirkham:
Am seeing this file `.ycm_extra_conf.py` and am wondering if it is needed

It appears to use CUDA 12.0 headers (not sure if that is expected)

https://github.com/rapidsai/pynvjitlink/blob/72a76ee544f0f6a8b3e28f7d1e8f1260175a609c/.ycm_extra_conf.py#L14

Also pins to Python 3.10 headers

https://github.com/rapidsai/pynvjitlink/blob/72a76ee544f0f6a8b3e28f7d1e8f1260175a609c/.ycm_extra_conf.py#L9

If this file is still needed, maybe it could benefit from some cleanup

If not, maybe we can just drop it

Note: It appears [the file is not used in the repo]( https://github.com/search?q=repo%3Arapidsai%2Fpynvjitlink+%22.ycm_extra_conf.py%22&type=code )

vyasr:
`.ycm_extra_conf.py` is a config file used by https://github.com/ycm-core/YouCompleteMe, a vim plugin for semantic completion and similar things. I'm guessing @gmarkall makes use of it for his local development since he committed the file initially, so I'm curious to hear if he wants to keep it for some reason.

jakirkham:
Thanks for the context! 🙏

Agree it would be great to hear from Graham 🙂

gmarkall:
I think it helps anyone using YCM. The hardcoded paths need to be fixed though, I'll submit a PR once I've finished tidying up #60.

jakirkham:
Thanks Graham! 🙏

In that case, what about something like this: https://github.com/rapidsai/pynvjitlink/pull/64 ?

Open to feedback there

jakirkham:
A more generalized approach is in PR: https://github.com/rapidsai/pynvjitlink/pull/65

Feel free to merge whichever works best 🙂

PR #64, opened 2024-03-06
Drop version hard-coding from YCM conf

jakirkham:
Fixes https://github.com/rapidsai/pynvjitlink/issues/63

* Dynamically determine the Python version
* Use `/usr/local/cuda` as this often points to the latest version

### DIFF
--- a/.ycm_extra_conf.py
+++ b/.ycm_extra_conf.py
@@ -5,13 +5,13 @@ from pathlib import Path
 
 CONDA_PREFIX = os.environ["CONDA_PREFIX"]
 NVJITLINK_DIR = Path(__file__).parent
-VER = sys.version_info
-PYTHON_INCLUDE_NAME = "python3.10"  # Not ideal
+PYTHON_VERSION = sys.version_info
+PYTHON_INCLUDE_NAME = "python%d.%d" % PYTHON_VERSION
 
 CONDA_INCLUDE_DIR = Path(CONDA_PREFIX, "include")
 PYTHON_INCLUDE_DIR = Path(CONDA_INCLUDE_DIR, PYTHON_INCLUDE_NAME)
 NVJITLINK_INCLUDE_DIR = Path(NVJITLINK_DIR, "include")
-CUDA_INCLUDE_DIR = "/usr/local/cuda-12.0/include"
+CUDA_INCLUDE_DIR = "/usr/local/cuda/include"
 
 flags = [
     "--cuda-gpu-arch=sm_50",


.ycm_extra_conf.py
```diff
@@ -5,13 +5,13 @@
 
 CONDA_PREFIX = os.environ["CONDA_PREFIX"]
 NVJITLINK_DIR = Path(__file__).parent
-VER = sys.version_info
-PYTHON_INCLUDE_NAME = "python3.10"  # Not ideal
+PYTHON_VERSION = sys.version_info
+PYTHON_INCLUDE_NAME = "python%d.%d" % PYTHON_VERSION
```
jakirkham, line 9: This could be a more robust solution

```suggestion
PYTHON_INCLUDE_NAME = sysconfig.get_config_var("INCLUDEPY")
```

.ycm_extra_conf.py
```diff
@@ -5,13 +5,13 @@
...
 
 CONDA_INCLUDE_DIR = Path(CONDA_PREFIX, "include")
```
jakirkham, line 11: Maybe we could generalize to something like this?

```suggestion
BASE_INCLUDE_DIR = sysconfig.get_config_var("INCLUDEDIR")
```

Would yield other changes around this

.ycm_extra_conf.py
```diff
@@ -5,13 +5,13 @@
...
 PYTHON_INCLUDE_DIR = Path(CONDA_INCLUDE_DIR, PYTHON_INCLUDE_NAME)
 NVJITLINK_INCLUDE_DIR = Path(NVJITLINK_DIR, "include")
-CUDA_INCLUDE_DIR = "/usr/local/cuda-12.0/include"
+CUDA_INCLUDE_DIR = "/usr/local/cuda/include"
```
jakirkham, line 14: Actually in the Conda case wonder if this is obviated by having the new CTK packages. Would be interested to know

gmarkall (approved):
These look good - many thanks!

.ycm_extra_conf.py
gmarkall: This would need to be assigned to `PYTHON_INCLUDE_DIR` below because it gives the full path.

.ycm_extra_conf.py
gmarkall: My use cases often include using non-released CUDA toolkits that I have installed in `/usr/local`, so it'd be preferable to stick with the `/usr/local/cuda/include` "default".

.ycm_extra_conf.py
gmarkall: I think we could make this change, but I'm not sure it would yield any other changes.

jakirkham:
Happy to go with this if we prefer

For comparison also tried using `sysconfig` in a separate PR: https://github.com/rapidsai/pynvjitlink/pull/65

No strong preferences. Whatever works best

gmarkall:
Let's go with this - it removes the hard-coding and works as-is.

jakirkham:
Sounds good. Thanks Graham! 🙏

PR #65, opened 2024-03-06
Rewrite of YCM conf with the `sysconfig` module

jakirkham:
Fixes https://github.com/rapidsai/pynvjitlink/issues/63

This is a rewrite of the YCM `conf` using the `sysconfig` module. It generalizes things a bit more and may work in other situations better

Would be interested to know if there are any tweaks needed (like for wheels)

### DIFF
--- a/.ycm_extra_conf.py
+++ b/.ycm_extra_conf.py
@@ -5,13 +5,13 @@ from pathlib import Path
 
 CONDA_PREFIX = os.environ["CONDA_PREFIX"]
 NVJITLINK_DIR = Path(__file__).parent
-VER = sys.version_info
-PYTHON_INCLUDE_NAME = "python3.10"  # Not ideal
+PYTHON_VERSION = sys.version_info
+PYTHON_INCLUDE_NAME = "python%d.%d" % PYTHON_VERSION
 
 CONDA_INCLUDE_DIR = Path(CONDA_PREFIX, "include")
 PYTHON_INCLUDE_DIR = Path(CONDA_INCLUDE_DIR, PYTHON_INCLUDE_NAME)
 NVJITLINK_INCLUDE_DIR = Path(NVJITLINK_DIR, "include")
-CUDA_INCLUDE_DIR = "/usr/local/cuda-12.0/include"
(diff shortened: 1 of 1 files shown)


.ycm_extra_conf.py
```diff
@@ -1,24 +1,18 @@
...
 NVJITLINK_DIR = Path(__file__).parent
-VER = sys.version_info
-PYTHON_INCLUDE_NAME = "python3.10"  # Not ideal
 
-CONDA_INCLUDE_DIR = Path(CONDA_PREFIX, "include")
-PYTHON_INCLUDE_DIR = Path(CONDA_INCLUDE_DIR, PYTHON_INCLUDE_NAME)
+PREFIX_INCLUDE_DIR = Path(PREFIX, "include")
+PYTHON_INCLUDE_DIR = sysconfig.get_config_var("include")
 NVJITLINK_INCLUDE_DIR = Path(NVJITLINK_DIR, "include")
-CUDA_INCLUDE_DIR = "/usr/local/cuda-12.0/include"
```
jakirkham, line 14: Figured search `PREFIX` would be good enough to pick these up

If not, we could add this back and use `/usr/local/cuda` here

jakirkham: Missed this comment: https://github.com/rapidsai/pynvjitlink/pull/64#discussion_r1515235864

Added back in commit: https://github.com/rapidsai/pynvjitlink/pull/65/commits/1cfd7115ada953ed57c1592e39eaf3d2bafbca12

.ycm_extra_conf.py
```diff
@@ -1,24 +1,20 @@
...
+CUDA_INCLUDE_DIR = "/usr/local/cuda/include"
 
 flags = [
     "--cuda-gpu-arch=sm_50",
```
bdice, line 13: RAPIDS requires arch 70 or higher.
```suggestion
    "--cuda-gpu-arch=sm_70",
```

bdice:
Can this be closed?

.ycm_extra_conf.py
jakirkham: Could you please submit a new issue or PR?

PR #69, opened 2024-04-15
Fix `__version__`

bdice:
This PR fixes the `pynvjitlink.__version__` attribute to read from the `VERSION` file.

Fixes https://github.com/rapidsai/pynvjitlink/issues/70

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.5.0
+    rev: v4.6.0
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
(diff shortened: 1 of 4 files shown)


leofang (approved):
Speedy fix! Thanks, Bradley!

jakirkham:
Thanks Bradley! 🙏

Would it be worth adding a test for packages as well? Just to make sure the `VERSION` file is getting included

bdice:
@jakirkham Currently `pynvjitlink` would fail to import if the `VERSION` is not included. I added a test that the `__version__` attribute exists and is a string with length greater than zero. I considered parsing the `__version__` in the test, but I didn't want to introduce a test dependency on `packaging`.

jakirkham (approved):
LGTM. Thanks Bradley! 🙏

PR #78, opened 2024-05-02
README: Don't use extra index for pip install

gmarkall:
pynvjitlink wheels appear to be available on PyPI, so we should recommend that as the canonical location to use.

See: https://pypi.org/project/pynvjitlink-cu12/#files

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ The [RAPIDS](https://rapids.ai) pynvjitlink library provides a Python binding fo
 ## Installation with pip
 
 ```shell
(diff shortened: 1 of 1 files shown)


jakirkham:
Thanks Graham! 🙏

Agree this is a good idea

Should we call out to the nightly index as well ( https://pypi.anaconda.org/rapidsai-wheels-nightly/simple )? Asking as that gets updated every time we tag here

PyPI is a bit more of a manual process around RAPIDS release time atm. So we tend to only update what we had working with that RAPIDS release when it ships. Not necessarily every tag

leofang (approved):

bdice (approved):

PR #84, opened 2024-06-03
Add xfail_strict to test suite

vyasr:
In light of https://github.com/rapidsai/pynvjitlink/issues/4#issuecomment-2145015880, figured we could start enforcing this programatically going forward.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,3 +33,6 @@ license-files = ["LICENSE"]
 provider = "scikit_build_core.metadata.regex"
 input = "pynvjitlink/VERSION"
 regex = "(?P<value>.*)"
+
+[tool.pytest.ini_options]
+xfail_strict = true


vyasr:
CC @gmarkall

gmarkall (approved):
I think this is a good idea. I hadn't realised it was an option, or I'd have set it before.

bdice (approved):
I pushed a small fix to the copyright check, and merged in `main` (since I think we require PRs to be up to date in this repo before merging -- at least that was my experience from the GitHub UI earlier today).

vyasr:
Ideally we'd remove that setting if so, that's fairly restrictive. Dev on this repo is slow enough (or rather, linear enough) that it might not matter much, though.

raydouglass:
> Ideally we'd remove that setting if so, that's fairly restrictive

This enforcement has been removed.

PR #87, opened 2024-06-04
[DO NOT MERGE] ref new codecovcli upload command

AyodeAwe:
This PR is primarily to confirm that the `codecovcli` invocation at least gets past the `Error loading Python lib '/tmp/_MEImwrRwT/libpython3.11.so.1.0': dlopen: /usr/lib/aarch64-linux-gnu/libm.so.6: version `GLIBC_2.35' not found (required by /tmp/_MEImwrRwT/libpython3.11.so.1.0)` reported [here](https://github.com/rapidsai/ci-imgs/pull/142#issue-2328451536).

These tests show that we're able to acheive this with the source distribution variant of the codecovcli (0.6.0) package.
- https://github.com/rapidsai/pynvjitlink/actions/runs/9371539018/job/25801234092?pr=87
- https://github.com/rapidsai/pynvjitlink/actions/runs/9371539018/job/25801234433?pr=87

A second test was carried out [in this job](https://github.com/rapidsai/literate-octo-potato/actions/runs/9370643165/job/25798101533?pr=827) which used the same installation format (source dist).

Accompanying `shared-workflows` PR is linked [here](https://github.com/rapidsai/shared-workflows/pull/215).

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -46,14 +46,14 @@ jobs:
       script: "ci/build_conda.sh"
       matrix_filter: ${{ needs.compute-matrix.outputs.BUILD_MATRIX }}
   test-conda:
+    secrets: inherit
     needs:
       - build-conda
       - compute-matrix
-    uses: rapidsai/shared-workflows/.github/workflows/conda-python-tests.yaml@branch-24.04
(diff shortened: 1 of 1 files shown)


bdice:
Thanks for testing this @AyodeAwe! I think it looks like it's working?

AyodeAwe:
> Thanks for testing this @AyodeAwe! I think it looks like it's working?

Yep

PR #90, opened 2024-06-12
WIP: make conda recipe data-loading stricter

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/72.

Proposes replacing a use of `.get()` in the conda recipe with `[]`, to ensure that we get a loud build error if the `"project"` table is not present when `conda-build` reads `pyproject.toml`.

## Notes for Reviewers

I know that at some time in the past, the `.get()` was introduced as a possible solution for errors observed at build time: https://github.com/rapidsai/pynvjitlink/pull/33#discussion_r1449525892

Pushing this to see if those are still an issue.

### DIFF
--- a/conda/recipes/pynvjitlink/meta.yaml
+++ b/conda/recipes/pynvjitlink/meta.yaml
@@ -5,7 +5,7 @@
     load_file="pynvjitlink/VERSION",
     regex_pattern="(?P<value>.*)"
 )[0] %}
-{% set project_data = data.get("project") %}
+{% set project_data = data["project"] %}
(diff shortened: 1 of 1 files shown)


bdice (approved):

jameslamb:
Oh I guess maybe `rapids-bot` is not enabled here? I don't have write access here, so whenever you have time could you merge this @bdice ?

bdice:
@jameslamb Can you follow up (with ops?) to fix this repo to use normal RAPIDS merge processes and get yourself permissions?

jameslamb:
yep, on it!

PR #92, opened 2024-06-27
Provide nvJitLinkVersion as a public API.

### LINKED ISSUE #38: Embed build-time nvjitlink version and make accessible
pynvjitlink should embed the CUDA version used at build-time as a compile-time constant accessible via a Python API. That would be very helpful for us for debugging in the future. Since we're statically linking our only recourse for debugging purposes right now is to check when a particular pynvjitlink package was built and then see what version of CUDA the build scripts used at that time.

_Originally posted by @vyasr in https://github.com/rapidsai/pynvjitlink/pull/33#discussion_r1449479441_

bdice:
Resolves #38 by making the result of `nvJitLinkVersion` a public API.

### DIFF
--- a/ci/install_latest_cuda_toolkit.sh
+++ b/ci/install_latest_cuda_toolkit.sh
@@ -13,4 +13,4 @@ if [ "${OS_ID}" != "rocky" ]; then
     exit 1
 fi
 
-yum install -y nvidia-driver cuda-nvcc-12-5 cuda-cudart-devel-12-5 cuda-driver-devel-12-5 libnvjitlink-devel-12-5
+yum install -y cuda-nvcc-12-5 cuda-cudart-devel-12-5 cuda-driver-devel-12-5 libnvjitlink-devel-12-5
--- a/ci/install_latest_cuda_toolkit.sh
+++ b/ci/install_latest_cuda_toolkit.sh
@@ -2,17 +2,15 @@
 # Copyright (c) 2024, NVIDIA CORPORATION
 
 # Installs the latest CUDA Toolkit.
-# Supports CentOS 7 and Rocky Linux 8.
+# Supports Rocky Linux 8.
 
 yum update -y
 yum install -y epel-release
 
 OS_ID=$(. /etc/os-release; echo $ID)
-if [ "${OS_ID}" == "rocky" ]; then
-    yum install -y nvidia-driver
-else
+if [ "${OS_ID}" != "rocky" ]; then
     echo "Error: OS not detected as Rocky Linux. Exiting."
     exit 1
 fi
 
-yum install -y cuda-toolkit-12-5
+yum install -y nvidia-driver cuda-nvcc-12-5 cuda-cudart-devel-12-5 cuda-driver-devel-12-5 libnvjitlink-devel-12-5
--- a/pynvjitlink/api.py
+++ b/pynvjitlink/api.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2023, NVIDIA CORPORATION.
+# Copyright (c) 2023-2024, NVIDIA CORPORATION.
 
 from enum import Enum
 from pynvjitlink import _nvjitlinklib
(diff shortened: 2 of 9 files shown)


ci/install_latest_cuda_toolkit.sh
```diff
@@ -2,17 +2,15 @@
...
-if [ "${OS_ID}" == "rocky" ]; then
-    yum install -y nvidia-driver
-else
+if [ "${OS_ID}" != "rocky" ]; then
     echo "Error: OS not detected as Rocky Linux. Exiting."
     exit 1
 fi
 
-yum install -y cuda-toolkit-12-5
+yum install -y nvidia-driver cuda-nvcc-12-5 cuda-cudart-devel-12-5 cuda-driver-devel-12-5 libnvjitlink-devel-12-5
```
bdice, line 16: This is a much more minimal set of packages and should significantly speed up wheel builds in CI. I had meant to make this change long ago and never took the time for it until I started trying to run CI locally.

pynvjitlink/api.py
```diff
@@ -16,6 +16,10 @@ class InputType(Enum):
     LIBRARY = 6
 
 
+def nvjitlink_version():
```
bdice, line 19: I'm open to naming suggestions if something else seems better.

gmarkall: I think this name is already good!

ci/install_latest_cuda_toolkit.sh
gmarkall: Thank you for cleaning this up :slightly_smiling_face:

ci/install_latest_cuda_toolkit.sh
```diff
@@ -2,17 +2,15 @@
...
 # Installs the latest CUDA Toolkit.
-# Supports CentOS 7 and Rocky Linux 8.
+# Supports Rocky Linux 8.
 
 yum update -y
 yum install -y epel-release
 
 OS_ID=$(. /etc/os-release; echo $ID)
-if [ "${OS_ID}" == "rocky" ]; then
-    yum install -y nvidia-driver
```
gmarkall, line 12: Was this important? Is its removal the reason why CI is failing?

gmarkall (changes requested):
The changes look good to me - many thanks for this!

I see there's something up with CI, but I think this will be good to approve / merge once CI is happy again.

ci/install_latest_cuda_toolkit.sh
bdice: I fixed this in 804afd4. We can build with driver stubs from `cuda-driver-devel-12-5` (to avoid installing a full driver in CI via `nvidia-driver`) but auditwheel complains about the driver stubs. I just told auditwheel to exclude `libcuda.so.1` and that's fine.

bdice:
@gmarkall CI is fixed up now. Can you approve?

gmarkall (approved):

pynvjitlink/api.py
leofang, line 19: I am wondering if this should be a (perhaps class) method instead? Such as `NvJitLinker.version()`?

gmarkall: I think that seems ambiguous or confusing, as it's providing the version of the underlying nvJitLink library, and `NvJitLinker.version` might appear to be the version of the pynvjitlink library.

bdice: I agree with Graham, it feels better as a free function named like “nvjitlink” and not “NvJitLinker.”

raydouglass (approved):

bdice:
I'm not sure why this isn't merging. I will try merging in the upstream and a few other things.

AyodeAwe (approved):

bdice:
The reason this wasn't merging was because the ops-bot requires the target branch to be something besides `main`. I opened a PR to address this: https://github.com/rapidsai/ops-bot/pull/188

PR #94, opened 2024-07-05
skip CMake 3.30.0

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/80

Adds constraints to avoid pulling in CMake 3.30.0, for the reasons described in that issue.


jameslamb:
The `test-conda` jobs here are fall failing like this:

```text
Error: Codecov token not found. Please provide Codecov token with -t flag.
```

([build link](https://github.com/rapidsai/pynvjitlink/actions/runs/9812827768/job/27097833390?pr=94))

I think that would be fixed by merging #92

PR #102, opened 2024-08-23
Drop Python 3.9 support

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/88.

Drops Python 3.9 support.

### Is this safe for `nvmath-python`?

I think yes.

In an offline conversation, I was told that `nvmath-python` follows NEP 29 (https://numpy.org/neps/nep-0029-deprecation_policy.html) and so would want to drop Python 3.9 in its next release anyway.

That's documented in its docs as well:

https://github.com/NVIDIA/nvmath-python/blob/7c485842d0f3300e03ec780056936503913910fe/docs/sphinx/overview.rst?plain=1#L274

cc @leofang 

### How I tested this

Checked that there were no remaining uses like this:

```shell
git grep -E '3\.9'
git grep '39'
git grep 'py39'
```

And similar for variations on Python 3.8 (to catch things that were missed the last time this was done).

### What are these other changes?

`ruff` raised two new errors here:

```text
.ycm_extra_conf.py:24:5: N802 Function name `Settings` should be lowercase
pynvjitlink/patch.py:226:20: UP031 Use format specifiers instead of percent format
Found 2 errors.
No fixes available (1 hidden fix can be enabled with the `--unsafe-fixes` option).
```

Left `Settings` untouched... looks like `ycm` expects that to be exactly `Settings` ([ycm-core docs](https://github.com/ycm-core/YouCompleteMe/blob/0d855962a60bd993980d287a3ac97f57ecd02ea6/README.md?plain=1#L1431)).

The `verify-copyright` hook also complained about one file missing a copyright header

```text
In file .ycm_extra_conf.py:1:1:
 import os
warning: no copyright notice found
```

So I added one.


bdice (approved):

jakirkham (dismissed):
Think Leo gets back next week. Let's wait for him to review

jameslamb:
Ah right. Ok no problem! This doesn't block the rest of RAPIDS, it can happen at any time.

leofang (approved):

leofang:
Thanks, James!

jakirkham (approved):

PR #106, opened 2024-10-10
make conda installs in CI stricter

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/106

Proposes specifying the package version in `conda install` calls that install CI artifacts, to reduce the risk of CI jobs picking up artifacts from other releases.


ci/test_conda.sh
```diff
@@ -5,6 +5,8 @@ set -euo pipefail
 
 . /opt/conda/etc/profile.d/conda.sh
 
+VERSION="$(head -1 pynvjitlink/VERSION)"
```
jameslamb, line 8: ref: https://github.com/rapidsai/pynvjitlink/blob/14639d8ffa6d67f6aa72a09991b9c125f277933b/conda/recipes/pynvjitlink/meta.yaml#L4-L7

bdice (approved):
Thanks @jameslamb. I think this is ready to merge.

jakirkham:
Thanks all! 🙏

Issue #21, opened 2023-11-28
[FEA] Expand supported python versions


brandon-b-miller:
**Is your feature request related to a problem? Please describe.**
Currently wheels are only built for python 3.9 and 3.10 because we rely on the rapids ci images, this limits non rapids consumers of pynvjitlink. 

**Describe the solution you'd like**
If we want to support other versions of python we'll need either our own images or new rapids images for this one purpose, or something else.

**Describe alternatives you've considered**
Initially we tried installing python manually via conda within each CI job allowing us to build with any python version, however this was messy and error prone

**Additional context**
N/A

weiji14:
Would it be possible to upload Python 3.12 wheels to https://pypi.nvidia.com/pynvjitlink-cu12? I'm trying to pip install `cudf-cu12=24.10` and am getting the following error:

```
error: distribution pynvjitlink-cu12==0.2.3 @ registry+https://pypi.nvidia.com/ can't be installed because it doesn't have a source distribution or wheel for the current platform
```

jakirkham:
Hi Wei, we moved the publication of this package to PyPI. There should be Python 3.12 packages over there. Please give it a try and let us know whether you still have issues

weiji14:
Ah yes, I see them at https://pypi.org/project/pynvjitlink-cu12/0.3.0/#files. I'll try and get my package manager to use that instead. Thanks!

jameslamb:
The most recent set of wheels for this project (https://pypi.org/project/pynvjitlink-cu12/0.7.0/#files) covered all of the same Python versions as RAPIDS (Python 3.10-3.13), and at times in the past this project has covered MORE versions of Python than RAPIDS was, via its own matrix:

https://github.com/rapidsai/pynvjitlink/blob/071cb4b4c5caa4f685e78e6f84d15b24e007867d/.github/actions/compute-matrix/action.yaml#L18-L38

So I think this can safely be closed, especially since the repo is about to be archived. See:

* #130 
* https://docs.rapids.ai/notices/rsn0052/

PR #108, opened 2024-10-21
Build with the Python limited API.

vyasr:
I noticed discussion on https://github.com/rapidsai/pynvjitlink/issues/21 around supported Python versions last week and it got me thinking about whether pynvjitlink really needs Python version-specific packages. pynvjitlink uses the C API directly, and it uses a pretty minimal subset, so we can actually just build it with the limited API and get a package compatible across Python versions. I tested the built wheel on 3.11 and 3.12 and tests pass just fine, at least locally.

The main blocker is that we require the limited API as of Python 3.11 (namely support for the buffer protocol) to make this switch. We _just_ dropped support for Python 3.9, so we're likely a year out from dropping 3.10. If we do think that the limited API is a path work exploring, our two options are to simply keep this PR around as a reminder until we drop 3.10, or to bifurcate our build into a 3.10 path and a 3.11+ path. I don't think the latter would be too painful to maintain. It would require three things that I can think of off the top of my head: 1) updating test jobs to download the abi3 wheel for Python>3.11 instead of one specific to their version (so in general, some of the improvements needed for https://github.com/rapidsai/build-planning/issues/43); 2) specifying `wheel.py-api` via a config-settings flag in build_wheel.sh for Python 3.10; and 3) updating the conda recipes accordingly to set the right run exports for the abi3 packages.

A middle ground might be to do nothing now, but the next time someone asks for support for Python 3.13 we switch to the limited API instead.

Since pynvjitlink has users beyond just RAPIDS, making the change here seems beneficial so that pynvjitlink can uncouple from the Python upgrade cycle of RAPIDS.

### DIFF
--- a/.github/actions/compute-matrix/action.yaml
+++ b/.github/actions/compute-matrix/action.yaml
@@ -16,19 +16,13 @@ runs:
         set -eo pipefail
 
         export BUILD_MATRIX="
-        - { CUDA_VER: '12.0.1', ARCH: 'amd64', PY_VER: '3.10', LINUX_VER: 'rockylinux8' }
         - { CUDA_VER: '12.0.1', ARCH: 'amd64', PY_VER: '3.11', LINUX_VER: 'rockylinux8' }
-        - { CUDA_VER: '12.0.1', ARCH: 'amd64', PY_VER: '3.12', LINUX_VER: 'rockylinux8' }
-        - { CUDA_VER: '12.0.1', ARCH: 'arm64', PY_VER: '3.10', LINUX_VER: 'rockylinux8' }
         - { CUDA_VER: '12.0.1', ARCH: 'arm64', PY_VER: '3.11', LINUX_VER: 'rockylinux8' }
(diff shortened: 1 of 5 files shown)


bdice:
I support this conceptually, but there are some other moving parts in the ecosystem. We would need to branch this into 3.10 and 3.11+ wheels. However, by the time we are ready to drop Python 3.10, I expect cuda-python to have nvjitlink bindings. We may have a shorter life for pynvjitlink as a package if we are able to adopt cuda-python's bindings. https://github.com/NVIDIA/cuda-python/issues/98

I am kind of leaning towards keeping things as-is and devoting effort towards adopting cuda-python's bindings once available.

jakirkham:
Would also note that Python 3.13 with free-threading enabled (so GIL disabled) [cannot use the Limited C API or Stable ABI]( https://docs.python.org/3/howto/free-threading-extensions.html#building-extensions-for-the-free-threaded-build ). So there likely will still be special casing with new Python versions

vyasr:
Yeah I was also thinking about the cuda-python story when I made this. I figured I would open the PR since it was easy to start the discussion, but I'm fine with assuming we'll be moving towards all of the logic eventually going into cuda-python + numba-cuda.

PR #111, opened 2024-10-29
WIP: Switch to use new Python bindings from cuda-python/cuda-bindings 

leofang:
Part of #130.

Note: Currently to try this PR it requires building cuda-python/cuda-bindings from source:
https://github.com/NVIDIA/cuda-python/tree/main/cuda_bindings

- Drop the old, CPython API-based nvJitLink bindings, in favor of the new one from CUDA Python
   - Support dynamic linking, no more rebuilding bindings for each CUDA minor release
      - nvidia-nvjitlink-cu12 wheel on Linux is not yet supported (on Windows it works), see https://github.com/NVIDIA/cuda-python/issues/98
   - Support Windows 
   - **pynvjitlink becomes a pure Python package**
- All tests in `pynvjitlink/tests/test_pynvjitlink.py` passed except one (see comment below)
   - I need someone to help take a look


pynvjitlink/tests/test_pynvjitlink.py
```diff
@@ -74,14 +73,15 @@ def test_get_error_log(undefined_extern_cubin, gpu_arch_flag):
...
+    with pytest.raises(_nvjitlinklib.nvJitLinkError):
+        # FIXME: For some reason this API would leak the error log to stderr
         _nvjitlinklib.complete(handle)
     error_log = _nvjitlinklib.get_error_log(handle)
     _nvjitlinklib.destroy(handle)
     assert (
         "Undefined reference to '_Z5undefff' "
         "in 'undefined_extern.cubin'" in error_log
-    )
+    ), f"{error_log=}"
```
leofang, line 84: This is the failing test:
```
$ pytest test_pynvjitlink.py  --maxfail=3
============================================================= test session starts =============================================================
platform linux -- Python 3.12.5, pytest-8.3.3, pluggy-1.5.0
rootdir: /local/home/leof/dev/pynvjitlink
configfile: pyproject.toml
collected 20 items                                                                                                                            

test_pynvjitlink.py ............F.......                                                                                                [100%]

================================================================== FAILURES ===================================================================
_____________________________________________________________ test_get_error_log ______________________________________________________________

undefined_extern_cubin = ('undefined_extern.cubin', b'\x7fELF\x02\x01\x013\x07\x00\x00\x00\x00\x00\x00\x00\x01\x00\xbe\x00~\x00\x00\x00\x00\x00...0\x00\x00\x00\x00\x00\x03\x00\x00\x00\x07\x00\x00\x18\x80\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00')
gpu_arch_flag = '-arch=sm_89'

    def test_get_error_log(undefined_extern_cubin, gpu_arch_flag):
        handle = _nvjitlinklib.create(gpu_arch_flag)
        filename, data = undefined_extern_cubin
        input_type = InputType.CUBIN.value
        _nvjitlinklib.add_data(handle, input_type, data, filename)
        with pytest.raises(_nvjitlinklib.nvJitLinkError):
            # FIXME: For some reason this API would leak the error log to stderr
            _nvjitlinklib.complete(handle)
        error_log = _nvjitlinklib.get_error_log(handle)
        _nvjitlinklib.destroy(handle)
>       assert (
            "Undefined reference to '_Z5undefff' "
            "in 'undefined_extern.cubin'" in error_log
        ), f"{error_log=}"
E       AssertionError: error_log='ERROR 9: finish\n\x00'
E       assert "Undefined reference to '_Z5undefff' in 'undefined_extern.cubin'" in 'ERROR 9: finish\n\x00'

test_pynvjitlink.py:81: AssertionError
------------------------------------------------------------ Captured stderr call -------------------------------------------------------------
error   : Undefined reference to '_Z5undefff' in '�����'
=========================================================== short test summary info ===========================================================
FAILED test_pynvjitlink.py::test_get_error_log - AssertionError: error_log='ERROR 9: finish\n\x00'
======================================================== 1 failed, 19 passed in 0.57s =========================================================
```

CMakeLists.txt
```diff
@@ -5,21 +5,7 @@ cmake_minimum_required(VERSION 3.26.4 FATAL_ERROR)
 project(
   pynvjitlink
   VERSION ${SKBUILD_PROJECT_VERSION}
-  LANGUAGES CXX CUDA
+  LANGUAGES CXX
 )
```
leofang, line 9: Given that this PR turns pynvjitlink into a pure Python package (I think!), I don't really know what to put here, given that I've never used CMake to package a pure Python project... 😆

pynvjitlink/tests/test_patch.py
leofang, line 1: For some reason I had to set the env var `NUMBA_CUDA_USE_NVIDIA_BINDING=1` to run the tests, otherwise I hit weird type errors when Numba tried to load the cubin etc using the Driver APIs (through ctypes).

pynvjitlink/tests/test_pynvjitlink.py
gmarkall: This might be a regression in the version of nvJitLink used. Is that on your local system? What is the nvJitLink version?

CMakeLists.txt
gmarkall: I don't think it makes sense to involve CMake for a pure Python package. Probably only a pyproject.toml would be necessary.

bdice: Yes, we would need to rearchitect the build systems and CI scripts to account for having only a single build of a pure Python package.

bdice:
@gmarkall @leofang If you'd like a hand with the repackaging to make pynvjitlink a pure Python package, let me know. I can help with some pieces of this.

pynvjitlink/_nvjitlinklib.py
```diff
@@ -0,0 +1,59 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.
+
+# A shim layer to use new bindings from CUDA Python
+from cuda.bindings import nvjitlink
+from cuda.bindings.nvjitlink import nvJitLinkError
+
+
+def nvjitlink_version():
```
warsawnv, line 8: This is kind of an odd looking module.  Is it provided just for backward compatibility?  Also, why is this one method not just called `version()`?

pynvjitlink/_nvjitlinklib.py
```diff
@@ -0,0 +1,59 @@
...
+def destroy(handle):
+    return nvjitlink.destroy(handle)
+
+
+def add_data(handle, input_type, data, filename):
+    nvjitlink.add_data(handle, input_type, data, len(data), filename)
+
+
+def add_file(*args, **kwargs):
+    raise NotImplementedError("seems unused by pynvjitlink")
```
warsawnv, line 25: If unimplemented, should this shim be omitted?  Is it better to get an `AttributeError` if someone tries to access this or a `NotImplementedError` when they do?

warsawnv (commented):
Just a couple more drive-by observations.  Feel free to ignore.

pynvjitlink/_nvjitlinklib.py
leofang: > This is kind of an odd looking module. Is it provided just for backward compatibility?

Yup! The whole purpose of this module is to kill the old one (that was implemented using CPython C APIs). This method replaces the following one:
https://github.com/rapidsai/pynvjitlink/blob/666375a8696c260b1431f5292e4413516af8e706/pynvjitlink/_nvjitlinklib.cpp#L58

pynvjitlink/_nvjitlinklib.py
leofang: It's been a while. Looks like I did a faithful translation 😂 
https://github.com/rapidsai/pynvjitlink/blob/666375a8696c260b1431f5292e4413516af8e706/pynvjitlink/_nvjitlinklib.cpp#L193-L197

gmarkall:
If test binaries need to be built for Windows, I added a script to numba-cuda that builds them - it might be useful to copy across to this PR when Windows support is added / tested: https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/tests/test_binary_generation/build.bat

leofang:
This is no longer needed now that https://github.com/NVIDIA/numba-cuda/pull/124 and https://github.com/NVIDIA/numba-cuda/pull/133 are resolved.

PR #112, opened 2024-10-29
use sccache, print sccache stats in builds

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/111

Proposes some small packaging/CI changes, matching similar changes being made across RAPIDS.

* printing `sccache` stats to CI logs
* updating to the latest `rapids-dependency-file-generator` (v1.16.0) and `rapidsai/pre-commit-hooks` (v0.4.0)
* always explicitly specifying `python` in calls to `rapids-upload-wheels-to-s3`

While doing that, I also noticed this project wasn't using `sccache`. This PR proposes using `sccache` to drive compilation, to speed up builds a bit.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
     hooks:
       - id: clang-format
   - repo: https://github.com/rapidsai/dependency-file-generator
-    rev: v1.13.11
+    rev: v1.16.0
     hooks:
       - id: rapids-dependency-file-generator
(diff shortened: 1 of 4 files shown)


bdice (approved):
There’s very little compilation for this package, so it’s fine if we have no sccache. Either way, I am approving this for uniformity’s sake.

jameslamb:
Alright yeah, thank you. I'll merge this as-is.

PR #121, opened 2025-01-23
Update to CUDA 12.8.0.

bdice:
Updates pynvjitlink to CUDA 12.8.0.


jakirkham:
Thanks Bradley! 🙏

Looks good 👍 

Do we want to release in this PR too?

bdice:
> Do we want to release in this PR too?

I'll file a separate PR for the release after this goes in.

robertmaynard (approved):

pynvjitlink/tests/test_pynvjitlink.py
```diff
@@ -84,7 +84,11 @@ def test_get_error_log(undefined_extern_cubin, gpu_arch_flag):
     )
 
 
-def test_get_info_log(device_functions_cubin, gpu_arch_flag):
+def test_get_info_log(device_functions_cubin, gpu_arch_flag, gpu_compute_capability):
+    if gpu_compute_capability < (7, 5):
+        pytest.skip(
```
vyasr, line 89: Should this be a skip or do we just need to capture warnings (or if it's just noisy logging, capture the output)?

bdice: I didn't want to do that because I thought it would take more effort than it was worth, and might increase the code size more than I wanted. If you have a concise way to implement what you're thinking, or if you think it's important, please feel free to push changes!

gmarkall (approved):

pynvjitlink/tests/test_pynvjitlink.py
vyasr: I'm fine with doing nothing here if it's not trivial.

PR #123, opened 2025-01-27
Add release process guide.

bdice:
This PR adds some docs for the release process.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,5 +49,28 @@ contributing to. Start with _Step 3_ from above, commenting on the issue to let
 others know you are working on it. If you have any questions related to the
 implementation of the issue, ask them in the issue instead of the PR.
 
+## Releases
+
+The release process for pynvjitlink involves the following steps:
+
+- Open a PR to update `pynvjitlink/VERSION` to the desired version.
+- Generate a short changelog with `git log v<PREVIOUS_VERSION>..HEAD --oneline --pretty=format:"- %s"`
+- Put the changelog in the version update PR description.
+- Once `main` is updated, tag the release:
+```
+git checkout main && git pull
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -49,5 +49,28 @@ contributing to. Start with _Step 3_ from above, commenting on the issue to let
...
+```
+- For the tag annotation, paste the same changelog as above, like this:
+```
+v<VERSION>
+
+- ... (bullet points on release items)
+```
+- Push the tag:
+```
+git push rapidsai v<VERSION>
```
vyasr, line 72: The first argument to push here isn't just rapidsai, it's the name of the remote corresponding to `https://github.com/rapidsai/pynvjitlink`. We should leave that out since I assume everyone uses different conventions. If you really want to make this easily copy-pasteable you could apply my suggestion, but either way the VERSION needs to be filled in so I don't think it's really worth the complexity. Maybe add a note based on my suggestion if you think that would help.
```suggestion
git push $(git remote -v | grep push | grep rapidsai/workflows | awk '{print $1}') v<VERSION>
```

vyasr (changes requested):
Thanks for this Bradley.

CONTRIBUTING.md
jakirkham: If we want to be more explicit, let's just use the URL

```suggestion
git push git@github.com:rapidsai/pynvjitlink.git v<VERSION>
```

That said, I trust that anyone following these steps can understand what is meant here

bdice: > I trust that anyone following these steps can understand what is meant here

Agreed. I like John's suggestion as it is slightly shorter.

vyasr (approved):

jakirkham (approved):

Issue #117, opened 2024-11-18
[FEA] numba-cuda compatibility


gmarkall:
Presently pynvjitlink patches and tests against Numba with its builtin CUDA target. The current release of [numba-cuda](https://github.com/nvidia/numba-cuda) incorporates the pynvjitlink patch so that patching is no longer required.

pynvjitlink should be updated so that:

- If numba-cuda is in use, then it does not patch its linker.
- If numba-cuda is in use, then it is at least version 0.0.18 (earlier versions don't include the patch, and may not be compatible with being patched by pynvjitlink).

This will also allow the Numba-related tests in the pynvjitlink CI to pass with later CUDA toolkits (e.g. 12.5) as it has fixed the expected PTX patterns generated by newer toolkits in the tests of atomics (See https://github.com/NVIDIA/numba-cuda/pull/40/files#diff-d48759455e529fd3994f3380640329c652dae73f5360cb977359c8c43461925b).

isVoid:
Can you clarify this issue? It seems like there's some conflict between point 1 and point 2. I remember for numba-cuda after 0.0.18, the patching from pynvjitlink no longer works because introduction of `ignore_nonlto` argument, and that we should use the builtin `nvjitlinker` from numba-cuda.

PR #129, opened 2025-02-18
Raise when pynvjitlink is in use with Numba-CUDA

isVoid:
This PR raises when numba_cuda is in use and user attempts to invoke `patch_numba_linker`. Instead, user should attempt to enable pynvjitlinker via `CUDA_ENABLE_PYNVJITLINK` config. I believe this should close https://github.com/rapidsai/pynvjitlink/issues/117

### DIFF
--- a/pynvjitlink/patch.py
+++ b/pynvjitlink/patch.py
@@ -55,7 +55,7 @@ if spec is not None:
 
     import numba_cuda
 
-    numba_cuda_ver = (int(x) for x in numba_cuda.__version__.split("."))
+    numba_cuda_ver = tuple(int(x) for x in numba_cuda.__version__.split("."))
     if numba_cuda_ver < (0, 2, 0):
         suggestion = "Instead, use NUMBA_CUDA_ENABLE_PYNVJITLINK environment variable to enable pynvjitlink features."
     else:
--- a/pynvjitlink/patch.py
+++ b/pynvjitlink/patch.py
@@ -60,6 +60,8 @@ if spec is not None:
         suggestion = "Instead, use NUMBA_CUDA_ENABLE_PYNVJITLINK environment variable to enable pynvjitlink features."
     else:
         suggestion = "Instead, use config.CUDA_ENABLE_PYNVJITLINK option to enable pynvjitlink features."
+
+    _numba_cuda_error += suggestion
 else:
     _numba_cuda_in_use = False
 
--- a/pynvjitlink/patch.py
+++ b/pynvjitlink/patch.py
@@ -9,12 +9,10 @@ from pynvjitlink.api import NvJitLinker, NvJitLinkError
 _numba_version_ok = False
 _numba_error = None
 
-_numba_cuda_version_ok = False
 _numba_cuda_in_use = False
 _numba_cuda_error = None
 
 required_numba_ver = (0, 58)
-max_numba_cuda_ver = (0, 0, 18)
 
 mvc_docs_url = (
     "https://numba.readthedocs.io/en/stable/cuda/" "minor_version_compatibility.html"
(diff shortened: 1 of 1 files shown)


pynvjitlink/patch.py
```diff
@@ -44,6 +48,24 @@
...
 
 
+spec = importlib.util.find_spec("numba_cuda")
+if spec is not None:
+    _numba_cuda_in_use = True
+    _numba_cuda_error = "`numba_cuda` includes patches from pynvjitlink, so no further patches are needed. "
+
+    import numba_cuda
+
+    numba_cuda_ver = (int(x) for x in numba_cuda.__version__.split("."))
```
gmarkall, line 58: On Python 3.11 at least, this results in comparing a generator with a tuple, which isn't supported:

```
In [1]: from pynvjitlink import patch
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[1], line 1
----> 1 from pynvjitlink import patch

File ~/numbadev/pynvjitlink/pynvjitlink/patch.py:59
     56 import numba_cuda
     58 numba_cuda_ver = (int(x) for x in numba_cuda.__version__.split("."))
---> 59 if numba_cuda_ver < (0, 2, 0):
     60     suggestion = "Instead, use NUMBA_CUDA_ENABLE_PYNVJITLINK environment variable to enable pynvjitlink features."
     61 else:

TypeError: '<' not supported between instances of 'generator' and 'tuple'
```

One way around this seems to be:

```suggestion
    numba_cuda_ver = tuple(int(x) for x in numba_cuda.__version__.split("."))
```

pynvjitlink/patch.py
```diff
@@ -44,6 +48,24 @@
...
+    if numba_cuda_ver < (0, 2, 0):
+        suggestion = "Instead, use NUMBA_CUDA_ENABLE_PYNVJITLINK environment variable to enable pynvjitlink features."
```
gmarkall, line 60: Can't you use `config.CUDA_ENABLE_PYNVJITLINK` for later versions as well?

pynvjitlink/patch.py
brandon-b-miller: I've had success with `looseversion` for this type of situation, might be worth depending on for ease of use and futureproofing.

pynvjitlink/patch.py
isVoid: Yes, but I think the support is incomplete - the user would have to do the config prior to the first import line of `numba.cuda`.

gmarkall: I think that was the case, but I think now as long as it's set before a kernel is jitted then it's OK - the config variable doesn't appear to be read until the linker is instantiated in `Linker.new()`.

gmarkall: Anyway, I think the recommendation is OK as-is, no need to change anything.

gmarkall (approved):

gmarkall:
@brandon-b-miller I don't foresee issues with cudf resulting from this, but want to check in with you - I think we're already using pynvjitlink "properly" in cudf anyway?

brandon-b-miller:
> @brandon-b-miller I don't foresee issues with cudf resulting from this, but want to check in with you - I think we're already using pynvjitlink "properly" in cudf anyway?

correct, [here](https://github.com/rapidsai/cudf/blob/566eb9156072570179c2df92dcfdfd46e50d7c88/python/cudf/cudf/utils/_numba.py#L136)

pynvjitlink/patch.py
isVoid: > I think that was the case, but I think now as long as it's set before a kernel is jitted then it's OK - the config variable doesn't appear to be read until the linker is instantiated in `Linker.new()`.

Ah, you are right about this. Sorry replied late on this.

bdice:
@gmarkall I'm not sure which suggestion you mean. Do you think this should merge or be closed?

brandon-b-miller:
My two cents here is that we shouldn't fuss too much with the pynvjitlink logic at this point with the imminent merge of https://github.com/NVIDIA/numba-cuda/pull/133. Rapids will likely skip most of the numba versions with these minor behavioral changes ultimately anyways and start with the next release that contains the linker changes. I don't feel strongly either way but I think the simplest thing would be to close.

gmarkall:
> @gmarkall I'm not sure which suggestion you mean. Do you think this should merge or be closed?

Sorry, I meant these suggestions to the user in the code no longer make sense:

```python
    if numba_cuda_ver < (0, 2, 0):
        suggestion = "Instead, use NUMBA_CUDA_ENABLE_PYNVJITLINK environment variable to enable pynvjitlink features."
    else:
        suggestion = "Instead, use config.CUDA_ENABLE_PYNVJITLINK option to enable pynvjitlink features."
```

gmarkall (approved):

PR #131, opened 2025-03-05
Add support for Python 3.13

bdice:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/120

This PR adds support for Python 3.13.

## Notes for Reviewers

This is part of ongoing work to add Python 3.13 support across RAPIDS.
It temporarily introduces a build/test matrix including Python 3.13, from https://github.com/rapidsai/shared-workflows/pull/268.

A follow-up PR will revert back to pointing at the `branch-25.04` branch of `shared-workflows` once all
RAPIDS repos have added Python 3.13 support.

### This will fail until all dependencies have been updates to Python 3.13

CI here is expected to fail until all of this project's upstream dependencies support Python 3.13.

This can be merged whenever all CI jobs are passing.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -21,10 +21,10 @@ jobs:
       - test-wheels
       - test-patch
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@branch-24.10
+    uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@python-3.13
   checks:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/checks.yaml@branch-24.10
+    uses: rapidsai/shared-workflows/.github/workflows/checks.yaml@python-3.13
     with:
       enable_check_generated_files: false
   compute-matrix:
@@ -40,7 +40,7 @@ jobs:
   build-conda:
     needs:
       - compute-matrix
-    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@branch-24.10
+    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@python-3.13
     with:
       build_type: pull-request
       script: "ci/build_conda.sh"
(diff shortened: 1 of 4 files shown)


gforsyth (approved):

.github/workflows/trigger-breaking-change-alert.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
   trigger-notifier:
     if: contains(github.event.pull_request.labels.*.name, 'breaking')
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/breaking-change-alert.yaml@branch-25.02
+    uses: rapidsai/shared-workflows/.github/workflows/breaking-change-alert.yaml@python-3.13
```
bdice, line 15:
```suggestion
    uses: rapidsai/shared-workflows/.github/workflows/breaking-change-alert.yaml@branch-25.04
```

dependencies.yaml
```diff
@@ -84,8 +84,12 @@ dependencies:
             packages:
               - python=3.12
           - matrix:
+              py: "3.13"
             packages:
-              - python>=3.10,<3.13
+              - python=3.13
+          - matrix:
+            packages:
+              - python>=3.10,<3.14
```
jakirkham, line 92:
```suggestion
              - python>=3.10,<3.14a0
```

bdice: @jakirkham Do you have thoughts on how to coordinate this suggestion with the proposal to drop all alpha specs? https://github.com/rapidsai/build-planning/issues/144

I am inclined to say we should leave this as-is, but I don’t have strong feelings.

.github/workflows/pr.yaml
```diff
@@ -21,10 +21,10 @@ jobs:
       - test-wheels
       - test-patch
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@branch-24.10
+    uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@branch-25.04
```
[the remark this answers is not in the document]
bdice: We have settled on doing these updates on an ad-hoc basis. I only changed them here because this PR was originally going to use the branch `python-3.13` but it turned out that isn’t necessary because the CI build/test matrix for pynvjitlink is independent of the shared-workflows matrix.

dependencies.yaml
jakirkham: Suppose we could have a different pin for conda & pip

That said, given one needs to opt-in to use RCs from conda-forge. It doesn't seem like that big of a risk

jakirkham (approved):

jakirkham:
Thanks Bradley! 🙏

PR #124, opened 2025-01-31
disallow fallback to Make in Python builds

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/146

Proposes:

* setting `[tool.scikit-build].ninja.make-fallback = false`, so `scikit-build-core` will not silently fallback to using GNU Make if `ninja` is not available

PR #132, opened 2025-03-07
Release 0.5.1

bdice:
This release updates some build and packaging scripts, and adds support for Python 3.13.

Includes:
- #123 
- #124 
- #125 
- #126 
- #131


jakirkham (approved):

vyasr (approved):

vyasr:
@webknjaz apologies for another direct ping, but this is sort of a follow up to the work you've seen us doing in e.g. https://github.com/NVIDIA/numba-cuda/pull/147 as we attempt to get everything using the workflow correctly. We've tried switching this repository to also use the GH runners to do a publish, but we're still seeing failures in https://github.com/rapidsai/pynvjitlink/actions/runs/13730908249/job/38407961979. We are indeed using a [GH runner now](https://github.com/rapidsai/pynvjitlink/actions/runs/13730908249/job/38407961979#step:1:1) (and [we're certainly requesting one](https://github.com/rapidsai/pynvjitlink/blob/main/.github/workflows/build.yaml#L17)). My assumption is that we're seeing failures because [we are using a container on top of the image](https://github.com/rapidsai/pynvjitlink/actions/runs/13730908249/job/38407961979#step:2:17), and I'm guessing that is also not supported. Is there any safe way for us to do this (e.g. by mounting something from the host into the container) or do we have to run bare metal on the runner?

webknjaz:
@vyasr that's unsupported and yes, that's likely because of the container. You could probably hack it somehow but I don't recommend it. Even if you somehow work around it, you'll hit another crash when you get to using Trusted publishing with digital attestations, because you're running it from a reusable workflow. It just won't work.
It is supposed to be the only thing running in the job, in the default runner. Nothing else is tested. And some things we can't even put into our CI.
As I said elsewhere, it must be in the top-level workflow. And the OIDC privilege would just be elevated for that job.

PR #137, opened 2025-04-22
feat(conda): port conda recipe to rattler-build

gforsyth:
Ports the condabuild recipe over to use rattler-build instead.

Contributes to rapidsai/build-planning#47
Contributes to rapidsai/build-planning#84

The only other major change here is that I am no longer passing in the
`cuda_compiler_version` variant variable using a separate file (`cat`ted in
from the `pynvjitlink/CUDA_VERSION` file. Instead, grabbing it from the
`RAPIDS_CUDA_VERSION` env-var. 

If we would prefer to read it from the file, I'll add that directly to the recipe, instead of side-loading it in the ci build script.


ci/build_conda.sh
```diff
@@ -15,23 +11,24 @@ export CMAKE_GENERATOR=Ninja
...
+# populates `RATTLER_CHANNELS` array and `RATTLER_ARGS` array
+source rapids-rattler-channel-string
 
-CUDA_VERSION="$(cat pynvjitlink/CUDA_VERSION)"
-export CUDA_VERSION
-
-cat > cuda_compiler_version.yaml << EOF
-cuda_compiler_version:
-  - "${CUDA_VERSION}"
-EOF
```
jakirkham, line 26: This was done to ensure `cuda_compiler_version` matched the intended `CUDA_VERSION` set in that file. That is the one canonical place used to CUDA version in this feedstock. It is also used by wheels to install the right Linux CUDA packages when building as well

Additionally setting `cuda_compiler_version` this way ensured it was considered when generating the package string hash. Maybe `rattler-build` handles this differently

conda/recipes/pynvjitlink/recipe.yaml
```diff
@@ -0,0 +1,71 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.
+schema_version: 1
+
+context:
+  cuda_version: ${{ (env.get("RAPIDS_CUDA_VERSION") | split("."))[:2] | join(".") }}
```
jakirkham, line 5: This project does things a bit differently than the rest of RAPIDS by using the `CUDA_VERSION` file noted above. Let's see if we can align on that somehow. Perhaps that can be done via setting an environment variable above

conda/recipes/pynvjitlink/recipe.yaml
```diff
@@ -0,0 +1,71 @@
...
+      SCCACHE_S3_KEY_PREFIX: pynvjitlink/${{ env.get("RAPIDS_CONDA_ARCH") }}/cuda${{ cuda_major }}
+
+requirements:
+  build:
+    - cmake >=3.24.4,!=3.30.0
+    - ninja
+    - ${{ compiler("c") }}
+    - ${{ compiler("cxx") }}
+    - ${{ compiler("cuda") }}
+    - ${{ stdlib("c") }}
```
jakirkham, line 50: Personally find it a little easier to have compilers first and then build dependencies. That way it is easy to see what language features are used. Then look at build tools after

conda/recipes/pynvjitlink/recipe.yaml
```diff
@@ -0,0 +1,71 @@
...
+  host:
+    - cuda-version =${{ cuda_version }}
+    - libnvjitlink-dev
+    - libnvjitlink-static
+    - python =${{ py_version }}
+    - pip
+    - rapids-build-backend >=0.3.0,<0.4.0dev0
+    - scikit-build-core >=0.10.0
+  run:
+    - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="x") }}
```
jakirkham, line 60: The upper bound needs to include minor version here. This has to do without `libnvjitlink` works in that it can only generate code that is usable by a matching CUDA version or older. It is also how we have it set in `meta.yaml`

https://github.com/rapidsai/pynvjitlink/blob/0cd61a92b6ff4ba5a5f5b9388a7fb06c6f7abbcc/conda/recipes/pynvjitlink/meta.yaml#L60

Additionally would suggest ordering `lower_bound` then `upper_bound`

```suggestion
    - ${{ pin_compatible("cuda-version", lower_bound="x", upper_bound="x.x") }}
```

jakirkham (commented):
Thanks Gil! 🙏

Took an initial pass below. There are some differences unique to this project compared to other RAPIDS projects highlighted below

Also do we want to delete the `meta.yaml` file as well?

ci/build_conda.sh
gforsyth: I can load it from the file directly and make sure it's part of the package string

gforsyth:
Thanks @jakirkham !  Applied your suggestions

conda/recipes/pynvjitlink/recipe.yaml
```diff
@@ -0,0 +1,69 @@
...
+      SCCACHE_S3_NO_CREDENTIALS: ${{ env.get("SCCACHE_S3_NO_CREDENTIALS") }}
+      SCCACHE_S3_KEY_PREFIX: pynvjitlink/${{ env.get("RAPIDS_CONDA_ARCH") }}/cuda${{ cuda_major }}
+
+requirements:
+  build:
+    - ${{ compiler("c") }}
+    - ${{ compiler("cxx") }}
+    - ${{ compiler("cuda") }}
+    - ${{ stdlib("c") }}
+    - cmake >=3.24.4,!=3.30.0
```
bdice, line 47: We bumped this for the rest of RAPIDS. Probably wouldn't hurt to match here since it's just a build-time requirement.
```suggestion
    - cmake >=3.30.4
```

bdice (approved):
Thanks @gforsyth and @jakirkham! I have one other comment, otherwise LGTM.

conda/recipes/pynvjitlink/recipe.yaml
```diff
@@ -0,0 +1,69 @@
...
+package:
+  name: pynvjitlink
+  version: ${{ version }}
+source:
+  path: ../../..
+
+build:
+  dynamic_linking:
+    overlinking_behavior: error
+  string: ${{ cuda_version }}_py${{ py_buildstring }}_${{ date_string }}_${{ head_rev }}
```
jakirkham, line 21: Can we add the package hash here too?

```suggestion
  string: ${{ cuda_version }}_py${{ py_buildstring }}_${{ date_string }}_${{ head_rev }}_h${{ PKG_HASH }}
```

ref: https://rattler.build/latest/build_script/#default-environment-variables-set-during-the-build-process

jakirkham (commented):
Thanks Gil! 🙏

This is looking pretty good. Had a comment below

Also think we need to move the recipe's `conda_build_config.yaml` to `variants.yaml`. Think the content can remain the same (though possibly minor tweaks will be needed)

conda/recipes/pynvjitlink/recipe.yaml
gforsyth: done and done!

